PR #31, opened 2016-11-09
Typo fix in page Title that appears in search engine results

Prudwell:


michaelrambeau:
@Prudwell Thank you very much for having found and fixed the typo.

How could I missed that typo, shame on me ;)

image

Issue #54, opened 2017-10-09
Add pagination for long lists of project 


michaelrambeau:
When displaying the "ALL PROJECTS" page (https://bestof.js.org/projects),  only the first 50 projects are displayed.

Same problem for tag pages that contain more than 50 projects, which only happens for the `React` tag right now.

For performance reasons, we cannot display all projects in one page, the application can become laggy if more than 100 project cards are displayed.

To face this problem we could implement a kind of infinite scrolling solution (using http://www.reactvirtualized.com/ for example) but the simplest solution is to implement a client-side pagination.

michaelrambeau:
## Screenshots

### On smaller screens

Only "Next Page" and "Previous Page" buttons are displayed

![image](https://user-images.githubusercontent.com/5546996/31323324-39264ed0-ace1-11e7-99e5-ba13c87d6fd3.png)

### On bigger screens

![image](https://user-images.githubusercontent.com/5546996/31323312-0d19cd3a-ace1-11e7-8275-c7252215dd82.png)

michaelrambeau:
Included in [0.10.0 release](https://github.com/michaelrambeau/bestofjs-webui/releases/tag/0.10.0)

PR #56, opened 2017-12-19
All projects view using `react-virtualized`

oz90210:


michaelrambeau:
Sorry @oz90210 , I don't know why you created a PR from a branch I created more than one year ago. This branch was only an experimentation.
I don't think we need `react-virtualized` right now because long pages like "ALL PROJECTS" are paginated, see #54.

PR #77, opened 2019-10-20
sort order: fix monthly label

blackpr:
fix the monthly label from `last 30 months` to `last month`

I'm not sure if this should be `last 30 days`


michaelrambeau:
Hello Tim @blackpr 
Thank you very much, good catch, it was an error indeed.

I think "the last 30 days" is more exact than "the last month" (especially we are in the middle of a month) but I'm merging your PR, I will adjust later.

PR #87, opened 2020-05-14
Check automatically for app update

michaelrambeau:
## Goal

When a new version of the application is deployed, we should ask users to update the application by reloading the browser page.

We could do do that by fetching every N minutes a very small file updated on the server after every request.

It could be a file called `meta.json` with the following content:

```json
{
  "date":"2020-05-14T00:49:23.867Z",
  "version":"0.23.0"
}
```


michaelrambeau:
## Screenshot

When a new update is detected, a notification shows us at the top right corner.

The user can either:

- "Update" to reload the window
- "Ignore", and the question will asked later again, the interval between every notification doubles every time the user skips the update.

image

PR #88, opened 2020-05-17
fix: Fixed the external link for wesbos.com

chiragswadia:
Currently the link to wesbos.com is broken as it does not have the protocol information

image


michaelrambeau:
Hello @chiragswadia 
Thank you for having found that bug.
I've just released a fix that adds `http` protocol to the URL, if it's missing.

This commit: https://github.com/bestofjs/bestofjs-webui/commit/c25f52b1bbabc4ab7ba8bff4ec20956f6f27c43f

Could you check if it works?

PR #98, opened 2020-08-13
Fix typo in old Babel name

desmondhume:
Babel used to be called `6to5`, not `5to6`.

Source: https://github.com/babel/babel/issues/780


michaelrambeau:
Thank you @desmondhume good catch! 👍

Issue #106, opened 2021-06-11
Dynamic titles for different pages


BBlackwo:
I realised today that the title is static (`Best of JS`) and doesn't get updated when you change pages.

It would improve SEO and share-ability if the title was changed based on the page you are on. E.g. for a specific project `Yarn | Best of JS`. Or for the hall of fame `JavaScript Hall of Fame | Best of JS`. This is what it looks like when you share a specific tag on Slack:

![Screenshot from 2021-06-11 15-16-59](https://user-images.githubusercontent.com/7598058/121634562-1fc1da00-cac8-11eb-9e15-be187972587a.png)

A first pass could be just use the `h1` of the page as the title, which would improve it a lot.

Ideally other metadata such as the `description` should be updated too.

If I get some time I'll try and implement this :slightly_smiling_face:

BBlackwo:
On a related note, the page titles should probably be `h1` instead of `h2` for better SEO/Accessibility.

michaelrambeau:
Thank you very much Benjamin, your are right the site is not optimized for search engines, one of the problems is that everything in generated in client, it could be nice to generate some pages on the server (the top page for example could be re-generated every day on the server when new data is available).

We need a "social image" too (the image that shows up when you share on social medias), like the nice one we did for the _Rising Stars_.

About the heading, in my mind "Best of JS" (the logo) was the top level title but it's just a logo so yes, let's change the HTML tags to start  the content of the page with a `<h1>` tag 👍 .

BBlackwo:
Yeah that would be great to do more server side rendering. That would be a big SEO improvement I imagine. Maybe 

A social image would be a nice improvement.

Changing to a `<h1>` sounds good :slightly_smiling_face:

PR #107, opened 2021-06-11
Add dynamic page titles using react-helmet

BBlackwo:
Add dynamic page titles using react-helmet as discussed in #106.

- Created `<Title>` that wraps the `<PageTitle>`
- In places where `<PageTitle>` is not used, use React Helmet directly


michaelrambeau:
Thank you Benjamin for your hard work, it works 👍 !

I checked online and I was happy to be able to see the title page, which is convenient when when you open a project's page for example.

image

Not very important but maybe we could merge the constants provided by this "container" https://github.com/bestofjs/bestofjs-webui/blob/develop/src/containers/static-content-container.tsx with the ones you created in the new file `constants/contants.ts`, especially because both mention the app title `Best of JS`.

But don't worry, you are not supposed to know the whole structure of the app (that has 6 years of history 😄!).

What I could do to decouple the development process and the release to production is to setup `develop` as the default branch (the one use to merge PR), while `master` is the production branch (that triggers the deployments to production).

This why I can merge your branch and later update the CHANGELOG and the version number that shows up in the "App Update" notification (i don't know if you have ever noticed this "toast").

michaelrambeau:
Thank you again for the improvements @BBlackwo Benjamin, I'm merging and I'm going to do a bit of refactoring because of the "technical debt".

BBlackwo:
Sounds good to merge the constants :+1: 

Good to know about the `develop` branch. Will use that next time.

I haven't seen the "App Update" notification :thinking: Does that just pop up when you load the page and it has updates?

michaelrambeau:
Hello @BBlackwo 

No worries, I have already setup the `develop` branch as the default one and I have merged everything (check the other PR #108 if you are interested)

The "toast" shows up when there is a release, it includes the next version number (if the user accepts and reloads the window).

image

PR #108, opened 2021-06-12
Simplify page title, add <PageHeader>

michaelrambeau:
## Goal

Following the great work initiated by @BBlackwo in #107,  simplify the page title handling, adding a new component `<PageHeader>` inspired by this one from AntD:

https://ant.design/components/page-header/

The component now renders a `h1` tag instead of `h2`.

#### `<PageHeader>` props

```ts
type PageHeaderProps = {
  title: string
  icon?: React.Node
  subTitle?: React.Node
  children?: ReactNode
}
```

Only `title` is required, it's a string because it's used both as the document title and the page main heading.
If for some reason we needed to show on the screen something more sophisticated we could use the optional `children` to pass any React node.

The other goal is to reduce the coupling to `react-helmet`library  as much as possible, to be able to switch to something else if needed (because it adds 5 KB to the bundle and we use it only for the page title!).
There is now only 1 occurrence of `react-helmet`.

Doing this we have the flexibility to decouple the document title and the heading on the screen.


michaelrambeau:
## Removing `react-helmet`

Actually we don't really need `react-helmet`, as `react-use` provides `useTitle` hook.

I created the following hook that implements the logic about the title suffix.

```tsx
import { useTitle } from 'react-use'

export function usePageTitle(title: string) {
  return useTitle(title + ' | Best of JS')
}
```

michaelrambeau:
## Styling adjustments

Decreasing the spacing around "Refine your search" section

#### Before

image


#### After

image

BBlackwo (commented):
Great cleanup! :rocket: Much tidier and good to remove a big library if we don't need it.

src/components/core/typography.tsx
```diff
@@ -23,42 +25,45 @@ export const ExternalLink = ({
...
   margin-right: 0.5rem;
   color: var(--iconColor);
   display: flex;
 `
 
-const PageTitleExtra = styled.span`
-  color: var(--textSecondary);
-  font-size: 16px;
-  margin-left: 0.25rem;
+const PageSubTitle = styled.span`
```
BBlackwo, line 65: I just noticed that this has low colour contrast (doesn't meet AA). We might wanna make it a bit darker?

Screenshot from 2021-06-15 13-27-44

michaelrambeau: Thank you Benjamin, i will fix that, maybe make it darker and smaller then, since I wanted the sub title to less visible than the title, like what Ant D does.

image

BBlackwo: That sounds good

michaelrambeau: Sorry for the late reply @BBlackwo

What do you think of this:

- same size
- darker color
- using a dot `•` as a separator

image

PR #116, opened 2021-09-28
Update about page links styling

BBlackwo:
## Goal

I noticed the links in the about page are not styled differently from the text. Now they are.

## How to test

Check out the about page :) 

## Screenshots

Screenshot from 2021-09-28 11-37-30


michaelrambeau:
Thank you Benjamin @BBlackwo good catch! 👍 

We use now a combination of 2 fonts:

- Roboto Slab for normal text
- Open sans for action and links (buttons and menu)

So I'm not sure what is best for simple links inside text 😄 (it's in the middle!)
Maybe it could be simplified by using one single font and that would improve performance a bit too.

BBlackwo:
Yeah we could try Open Sans for everything? :)

PR #137, opened 2022-05-20
Make sure timeline page not broken

nvh95:
## Goal
- Fix broken page https://bestofjs.org/timeline 
cannot access timeline page

After some investigation, the reason this page is broken is the data in `src/components/timeline/featured-projects.json`, which has `angular-1` and `vuejs`. But data get from https://bestofjs-static-api.vercel.app/projects.json (`state.entities.projects`) does not have `angular-1` and `vuejs` (but have `angular`, `vuejs-2`, `vuejs-3`). Then, logic in `useTimelineProjects` make the page broken.

```js
function useTimelineProjects() {
  const projects: BestOfJS.Project[] = useSelector(
    findProjectsByIds(featuredProjects.map(({ slug }) => slug))
  ); // CONTAIN `null`
  return featuredProjects.map(({ slug, date, comments }) => {
    const project: BestOfJS.Project | undefined = projects.find(
      (project) => project.slug === slug // CANNOT FIND slug of null
    );
    if (!project) return null;
    return {
      comments: comments || [],
      date: date || new Date(project.created_at),
      ...project,
    };
  }); // CONTAIN null project
}
```

This PR tries to make the page not broken. It just doesn't render the project which is missing data. However, the root cause is the data from the server is missing/ inconsistent. We still need to update the data since data inconsistency is the root cause. (But unluckily, I haven't found where to update, please help with that).

## How to test
- `npm run start` then head to `http://localhost:3000/timeline`. We should be able to see the page rendered (without Angular and Vue, have 18/20 projects).

## Screenshots
See above

### DIFF
--- a/src/components/timeline/timeline.tsx
+++ b/src/components/timeline/timeline.tsx
@@ -22,18 +22,20 @@ const template = tinytime("{MMMM} {YYYY}");
 function useTimelineProjects() {
   const projects: BestOfJS.Project[] = useSelector(
     findProjectsByIds(featuredProjects.map(({ slug }) => slug))
-  );
-  return featuredProjects.map(({ slug, date, comments }) => {
-    const project: BestOfJS.Project | undefined = projects.find(
-      (project) => project.slug === slug
-    );
-    if (!project) return null;
-    return {
-      comments: comments || [],
-      date: date || new Date(project.created_at),
-      ...project,
-    };
-  });
+  ).filter(Boolean);
+  return featuredProjects
+    .map(({ slug, date, comments }) => {
+      const project: BestOfJS.Project | undefined = projects.find(
+        (project) => project.slug === slug
+      );
+      if (!project) return null;
+      return {
+        comments: comments || [],
+        date: date || new Date(project.created_at),
+        ...project,
+      };
+    })
+    .filter(Boolean);
 }
 
 export const Timeline = () => {


nvh95:
@michaelrambeau Could you please look at this? https://bestofjs.org/timeline is currently broken in production. Thanks.

michaelrambeau (approved):
👍 Thank you very much for the fix @nvh95 

I'm going to deploy to production as soon as possible.

I will have to find a solution for the missing projects.

michaelrambeau:
@nvh95 it's live, thank you again!

Among the other things I have to fix,  we need a better error handling (using "ErrorBoundary") to avoid showing the default error page (that sometimes shows up after the site is updated, because of code splitting).

I will take a look at this and "restore" the missing projects.

nvh95:
@michaelrambeau Ah it makes sense to me now. Let me see if I can improve it.
The idea is still show the default error page, but with a box contains more error info (which can toggle on or off?). So it's easier to locate the bug faster.

michaelrambeau:
@nvh95  About the error handling, the main thing is to avoid breaking the whole user experience if something minor breaks.

For now there is only one level of error handling, it's better than nothing because we show an error page with a message, and navigation links to let the user "escape".

If there was no error handling, it would be just a white page with nothing displayed since everything is displayed.

However the current error handling is not great, because some parts in the app are more fragile but should not lead to the error page if they break.
For example in the top page, there is a kind of carousel that shows the featured projects.
If it breaks, we don't want to show the whole error page, we only want to either remove the broken part of show something that says the feature is not available.

Actually https://remix.run/ top page explains well how applications should be **resilient**, check their feature called "route error boundaries".

But don't worry if you don't have time, there are so many things to improve 😄 !

michaelrambeau:
I have just rebuilt _Best of JS_ data to include the projects that were missing.

image

nvh95:
@michaelrambeau 
> However the current error handling is not great, because some parts in the app are more fragile but should not lead to the error page if they break.

We can add more Error Boundaries to the fragile parts of the app, with customize `fallback` UI in each part:
image

https://github.com/bestofjs/bestofjs-webui/blob/bb662931fe8e965e74575ebb036727436f85cd2f/src/components/home/home.tsx#L70

Do you have a list of fragile parts in the app. I can help to add more Error Boundaries there.

Remix with `react-router` v6 is very promising. But I think that's the story of the near future, not immediately.

michaelrambeau:
Thank you for your help on this topic too @nvh95 

The first step is to identify what is the **shell** of the application and not what are the **extra** features, non essential, especially the ones that seem to be more fragile.

The goal is: if an extra feature breaks, we don't want the shell to be broken.

To me the shell is made of the features implemented by processing the huge JSON data: basically all project and tag lists, including search features.

We should start with the most important page, the top page:

- shell: "Hot Projects", "Popular Tags", "Recently Added"
- extra: "Featured Projects", "Monthly Rankings"

And I'd start by ensuring that the "featured projects" block can break without damaging the user experience because it's has a combination of dangerous elements: timeouts, animations, intersection, observer (used to detect when the element is not visible).

Another interesting page to check is the "Project Details" page because there are 2 extra API calls to get project statistics and README data. I think there is already some error handling related to the API calls but it would be safer to ensure that the page's  shell (basically the block at the top with the project links) can be displayed even if an error occurs when processing data.

The next step is to go "one lever deeper" and to investigate for any given element, what is essential, what can break... but I'm not sure we need to go so far, we may have other priorities.

Note: I mentioned Remix because they explain well the concept on their home page, I was not advocating using their router in our current application.

PR #139, opened 2022-05-22
Format dates with browser native `Intl.DateTimeFormat`?

michaelrambeau:
## Goal

Use browser's native `Intl.DateTimeFormat` API to format dates instead of using `date-fns`s `format` function, following the approach described here: https://dev.to/dmtrkovalenko/you-might-not-need-date-fns-23f7

This way we could remove `date-fns` dependency.

This is an experiment as I have never used this browser API before.

## How to test

Check the dates at 2 following places:

- At the footer of all page, the time data was updated: `The last update was 3 hours ago (at 6:16 AM).`
- In the `/timeline` page, check the dates of creation for each project (E.g. `August 2006`)

### Footer

image

### Timeline

image


michaelrambeau:
## Impact on the bundle sizes

Comparing the sizes of the bundles in the browser, this change would only save 6 kB.

- Before https://bestofjs-git-develop-bestofjs.vercel.app/ `180.05`
- After https://bestofjs-git-try-native-intl-date-time-format-bestofjs.vercel.app/ `174.13`.

So I'm not sure it's the best move 😄 !

nvh95 (changes requested):
LGTM!
For the current use case of Best Of JS. I think leveraging the browser's native API is the best move. It also has [excellent support (even supports IE11)](https://caniuse.com/?search=Intl.DateTimeFormat).
In the future, if the use case is getting more complex, I think we can consider to re-install `date-fns`.

Just one comment: Please remove `date-fns` in the `package.json` and `package-lock.json` as well

nvh95 (approved):

PR #140, opened 2022-05-22
Remove unused config for ejs

nvh95:
## Goal
- Remove redundant `ejs` config in `vite.config.ts`.
- I thought that config `inject.data.env` is to enject to `process.env`, but it's not correct. accessing `<% if (process.env.VERCEL_ENV === 'production') { %>` is supported by `ejs` natively.
- If I config as
```js
inject: {
  data: {
    env,
  },
},
```
I should write in the `index.html` like this
```js
<% if (env.VERCEL_ENV === 'production') { %>
```
But I use `process.env.VERCEL_ENV`, so no need any injection.

## How to test
Change 
- Build the app in PROD mode
```command
VERCEL_ENV=production npm run build
```
=> We should see the cloudflare analytics in the index.html

- Build the app in Preview mode
```command
VERCEL_ENV=preview npm run build
```
=> We should not see the cloudflare analytics in the index.html

cc: @michaelrambeau

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -18,11 +18,6 @@ export default defineConfig((props) => {
       react(),
       tsconfigPaths(),
       createHtmlPlugin({
-        inject: {
-          data: {
-            env,
-          },
-        },
         minify: true,
       }),
     ].filter(Boolean),


michaelrambeau (approved):
Thank you very much @nvh95 

The simpler, the better 😄 

According to some tests I did yesterday, it works, we are able to use directly variables provided by Vercel platform (E.g. `VERCEL_ENV` that is equal to `production` or `preview`, depending on the branch)

Just for the records, the test I did I:
- I added the following code to the HTML template
- I pushed a branch, deployed on Vercel

```ejs
 <% if (process.env.VERCEL_ENV === 'preview') { %>
    <script>
      console.log("preview");
    </script>
 <% } %>
```

PR #138, opened 2022-05-20
Increase Developer Experience by replacing Create React App with Vite

nvh95:
## Motivation
- Decrease the start and build time significantly. Start under 500ms, build under 5 seconds.

## Benchmark

|      | npm start | npm build |
|------|-----------|-----------|
| CRA  | 8.2s      | 17s       |
| Vite | 0.4s      | 5.1s      |

Vite is **20 times faster** when starting a dev server.
Vite is **3.4 times faster** when building app.
(Benchmarked on Macbook Air M1 2020)

## Changes
- [x] Packages changes
  - [x] Remove `react-scripts`
  - [x] remove `react-router`, only need react-router-dom
  - [x] Remove `tinytime`. Since it has some issue with Vite (`jsnext:main` point to `.js` code with flow syntax makes Vite confused). Replaced by `date-fns` (6kb gzipped)
  - [x] Add `fs-extra`. Currently, it’s not in the `package.json`, but the dependency of `react-scripts`
  - [x] Update `typescript` to newest version (`4.6.3`)
- [x] Fix some typescript issues
- [x] Remove unused CRA service worker code
- [x] Fix a typo in routes (`src/pages/routes.tsx`):
```diff
-<Route from={`/tags`} component={TagsPage} />
+<Route path={`/tags`} component={TagsPage} />
```
- [ ] Variable environments
  - [x] Change prefix: `REACT_APP` => `VITE_APP`. 
  - [ ] Need to update in Vercel settings as well (if have any) @michaelrambeau to do and mark as done
  - [ ]  Remeber to update `REACT_APP_ENV: production` to `VITE_APP_ENV: production` in Vercel as well @michaelrambeau to do and mark as done
- [x] Support [ejs](https://ejs.co/) in `index.html` to inject `cloudflareinsights` script.
- [x] Support light mode and dark mode logo (`bestofjs-logo.svg`) (#E65100 vs ##F6AD55)
- [x] Integrate Jest with Rust-based compiler SWC
- [x] Add GitHub Actions to run tests on Push and PR

## How to test
- Just run:
```bash
npm i
npm run start
```

## Screenshots
dev sever ready less than 0.5 seconds

PR #141, opened 2022-05-23
Add tests for `Best Of Js` (`msw` + `@testing-library/react` +`jest-preview`)

nvh95:
## Changes

**Setup and code changes:**

- [x] Add [msw](https://mswjs.io/) to mock APIs
  - [x] In browser: Run `npm run start:mock`
  - [x] In Node (tests): Run `npm run test`
- [x] Add and configure [Jest Preview](http://www.jest-preview.com/) to debug and write tests faster
- [x] Update test script to include environment variables
- [x] Add `data-testid` to use with `@testing-library/react`
- [x] Use `@ts-expect-error` instead of `@ts-ignore`
- [x] Add [eslint](https://eslint.org/) (accidentally remove it along with CRA in #138 )
- [x] Improve a11y on navigation buttons
- [x] Add a TODO to handle sign out refreshing issue

**Testing progress:**

- [x] Homepage
- [x] StaticPages (Timeline and About)
- [x] Authentication
- [x] DarkMode
- [x] MonthlyRank
- [x] Projects
- [x] Tags

**TODO in next PRs:**
  - Add a coverage service

## How to test
- Local: `npm run test`
- CI (GitHub Actions): `npm run test:ci` 

## Screenshots

jest test result

Jest total running time

Coverage report


nvh95:
This is how I write tests using GitHub Copilot 😀

https://user-images.githubusercontent.com/8603085/170552984-923017d4-4677-464c-a7e6-7b14e38ac72b.mp4

src/components/project-list/project-table.tsx
```diff
@@ -68,7 +75,7 @@ const ProjectTableRow = ({
   };
 
   return (
-    <Row>
+    <Row data-testid="project-card">
```
michaelrambeau, line 78: I thought the `data-test-id` attibute had to be unique in the page?
Table can have multiple rows.

src/components/project-list/project-table.tsx
```diff
@@ -20,10 +20,17 @@ type Props = {
   footer?: React.ReactNode;
   showDetails?: boolean;
   metricsCell?: (project: BestOfJS.Project) => React.ReactNode;
+  testid?: string;
```
michaelrambeau, line 23: Maybe I'm wrong but I find a bit annoying that we change the API of this component to add a property used only for testing.

Do you think we really need it? Usually table of projects are just below a title, could we target the tables using the titles?

E.g. if we need to target the first table of projects in the top page, something like "the table just after the title `Hot projects`?

About the test IDs the [documentation](https://testing-library.com/docs/queries/about#priority) says:

> getByTestId: The user cannot see (or hear) these, so this is only recommended for cases where you can't match by role or text or it doesn't make sense (e.g. the text is dynamic).

src/components/project-list/project-table.tsx
nvh95: @michaelrambeau no, `data-testid` does not need to be unique, `id` does.
We can actually query like this 
```js
const allProjects = screen.findAllByTestId("project-card");
```

src/components/project-list/project-table.tsx
nvh95: @michaelrambeau I agree. Let me try to improve it.

src/components/home/home.tsx
```diff
@@ -68,18 +68,20 @@ export const Home = ({
...
-              />
-              <CompactTagList
-                tags={popularTags}
-                footer={
-                  <Button as={RouterLink} to={`/tags/`} variant="link">
-                    View all tags »
-                  </Button>
-                }
-              />
+              <div data-testid="popular-tags-section-desktop">
```
michaelrambeau, line 71: Are we are changing the DOM structure, adding a new `<div>` only for testing purpose?

Besides the fact that adding an extra div might have CSS effects, usually I try to remove `div` that are not necessary because every DOM element has a cost and it makes inspecting things in DevTools more complicated.

However we may need a wrapper for these 2 component that are showed/hidden together depending on the screen size, we may need some refactoring, so your approach may be correct too, I need to think more...

nvh95: > Are we are changing the DOM structure, adding a new \<div\> only for testing purpose?

Yes. I need to change DOM so I can easily query (2).
popular tags in homepage

Besides, I strongly think (1) and (2) should be inside a wrapper, I believe that would be a better HTML structure if `Featured Projects` and `Popular Tags` are on the same HTML level. (Now `Featured Projects`, the heading `Popular Tags` and the list are on the same level)

side bar code

https://github.com/bestofjs/bestofjs-webui/blob/bb662931fe8e965e74575ebb036727436f85cd2f/src/components/home/home.tsx#L70-L82

> Besides the fact that adding an extra div might have CSS effects

I did think about if it might affect CSS. I tested it myself and look like it's going to be fine. Please help to test it by yourself.

> usually I try to remove div that are not necessary because every DOM element has a cost and it makes inspecting things in DevTools more complicated.

I 100% agree, but without changing the DOM, I couldn't test the `Popular Tags` part. So I guess this is the change we need to make. Please propose any other suggestion, since you are more familiar with Best Of JS code base.

src/components/project-list/project-table.tsx
nvh95: I removed the `testid` prop. Instead, I use the `hot-projects-section` selector.

src/index.tsx
```diff
@@ -11,4 +11,16 @@ function start() {
   render(<Root />, document.getElementById("root"));
 }
 
-start();
+const prepare = async (): Promise<void> => {
+  if (process.env.VITE_APP_MOCK === "1") {
+    console.log("a");
```
michaelrambeau, line 16: It'a good idea to print something as we are starting the "mock" mode but maybe we should show something more explicit like "Running with mock data, all HTTP calls will be intercepted by MSW" (if my understanding is correct?)

Spa

public/mockServiceWorker.js
```diff
@@ -0,0 +1,338 @@
+/* eslint-disable */
+/* tslint:disable */
+
+/**
+ * Mock Service Worker (0.41.0).
+ * @see https://github.com/mswjs/msw
+ * - Please do NOT modify this file.
+ * - Please do NOT serve this file on production.
```
michaelrambeau, line 8: I wonder why we are supposed to not touch this file, is it boilerplate configuration (where does it come from?), you have not changed anything?

src/index.tsx
nvh95: @michaelrambeau My bad when I leave the log I used while developing on this PR. I'm going to remove it soon 😂😂😂😂 
`msw` itself already provides that log when we run `npm run start:mock`

public/mockServiceWorker.js
nvh95: @michaelrambeau Yep. Do not touch this file. It is generated by running `npx msw init public/ --save`
This file is rarely changed, but whenever a new version of `msw` requires changes in this file, they will update it automatically when we execute `npm install msw`/ `npm update msw`

Reference: https://mswjs.io/docs/getting-started/integrate/browser#setup

michaelrambeau: Thank you for the explanation 👍

michaelrambeau:
It's not a big deal but I think mocks data would be simpler to understand if each response has its own file, whose name reflects the original end-point, because `index.ts` inside `mocks` folder is pretty big.

- `project-details-xxxx.json` (depending on the project you picked) for the project details data
- `project-readme-xxx.html` for the README request
- `monthly-rankings-2022-04.json` for the monthly rankings and so on...
 
Also:

project.json => projects.json (plural, it's a huge list!)

This way the mock response file can also serve as a kind of documentation about the API we use.

Sorry for asking so many changes that!

nvh95:
@michaelrambeau The idea of using file names as documentation is just brilliant. I pushed the changes. Please take a look.
Also, I have to name the file as `project-readme-xxx.html.js` if I save it as `html` file. I need to use `fs.readFileSync`, which is a Node-specific function, and it broke the app in the browser environment.

nvh95:
Bonus: Using Jest Preview saves me a ton of time when writing those tests and debugging them!!!!!

src/mocks/data/user-profile.json
```diff
@@ -0,0 +1,44 @@
+{
+  "bio": "Developing Jest Preview.",
+  "blog": "https://hung.dev",
+  "company": "Looking for job",
```
michaelrambeau, line 4: It's cool, we have a snapshot of your profile riight in _Best of JS_ 😄 

I'm pretty sure there you will belong to a great company soon!

nvh95:
@kettanaito Thanks for accepting my invitation to review this PR. It's our honor to have @mswjs's author take a look at our PR.

We have @michaelrambeau, author of @bestofjs, he created @bestofjs in 2015 with the mission of helping everyone keep track of the evolution of the Javascript ecosystem.

To make the best of your time here, I can brief what is happening here, with some concise instructions. Then we ask you to take a look at `msw`-related code of this PR (if you want to give feedback on other parts, it's more than welcome).

Basically, I used `msw` to mock all the APIs, for 2 environments:
1. **Browser:** Run when developing new feature `npm run start:mock`. 
2. **Node:** Re-use mocks to run tests.

The workflow will be: **Mock all APIs** => **Develop features** => **Reuse all the mocks to write integration tests** in Jest, with the support of `@testing-library/react` (render and select element), `@testing-library/user-event` (interactive with JSDOM), and `jest-preview` (preview UI in Jest to the browser, to write new tests and debug faster)

We have `mocks` folder contains almost `msw`-related code
- `mocks/api`: definitions of all mocked APIs
- `mocks/data`: all the mock data
- `src/index.tsx`: I use dynamic import instead of `require` to avoid `msw` code getting bundled to the production build (I broke production once when version `0.36.0` has a bug on safari https://github.com/mswjs/msw/issues/1018)

For other parts, I just follow your instruction on https://mswjs.io/

To run this app, you can use
```bash
npm run start          // to run the app normally
npm run start:mock     // to run the app with `msw` service worker enabled
npm run test           // to run Jest tests, mocked by `msw`
npm run jest-preview   // to preview the tests on browser (optional)
```

That's it. Thank you very much for your help. I am looking forward to your precious feedback.

michaelrambeau (approved):
@nvh95 👍 Thank you again Hung, this is an amazing amount of work, especially for someone who is new to the code base.

It's pretty impressive to have coverage data when tests runs on GitHub actions (even if I think there is no need to chase the perfect score, given the fact the application architecture may change soon). 

The test suite will help us if we do major updates such as package upgrades (I was thinking to migrate to ChakraUI 2 that released a few days ago, for example)

I will try to enjoy the benefits of Jest Preview when debugging tests very soon.

Thank you for having handled my feedback about the mock data, it gives us a very good overview of data retrieved by the application!

PR #146, opened 2022-06-15
Sanitize input before rendering to users

nvh95:
## Goal
- [x] Sanitize input before rendering to users using `dangerouslySetInnerHTML` (`src/components/project-details/readme-card.tsx`)
- [x] Run prettier on the whole project

## Why
We add one more layer of security so if our database is compromised. Our users are not at stake. This is [Defense in Depth](https://www.forcepoint.com/cyber-edu/defense-depth) principle (if one layer fails, another steps up immediately to thwart an attack)

## How to test

- Update `src/mocks/data/project-readme-react.html.js` like this

```diff
const projectReadmeReact = `<div id="readme" class="md" data-path="README.md">
+<img src/onerror=alert('stole-your-token')>
  <article class="markdown-body entry-content container-lg" itemprop="text">
...
```

- Run
```
npm run start:mock
```

- Navigate to any project, for example: http://localhost:3000/projects/freecodecamp. You should see the javascript is executed if we do not `sanitize` input.

## Screenshots

pop up in an alert window


michaelrambeau (approved):
@nvh95 Thank you very much for caring about the security of our users!

I was able to reproduce the issue you mentioned, about the use of JS code inside the HTML 👍 .

Also I can see the problem of end-of-line is now fixed mentioned previously https://github.com/bestofjs/bestofjs-webui/pull/145#issuecomment-1151760732 is fixed, we have `CR` everywhere, thank you!

```
➜  bestofjs-webui git:(sanitize-input) npx prettier --check src/* 
Checking formatting...
All matched files use Prettier code style!
```

About the bundle, since we are doing code splitting, fortunately there is mo change about the main bundle used to display the top page.

For the records, I'm sharing data about he project "details" page, from `npm run build' command:

- before: `18.09 KiB / gzip: 6.46 KiB`
- after: `38.61 KiB / gzip: 14.46 KiB`
 
So `dompurify` library adds only 8 KiB to the bundle, which is consistent with data we can see here: https://bundlephobia.com/package/dompurify@2.3.8 
 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

nvh95:
@michaelrambeau FYI that we also have alternatives to `dompurify` as @slorber's [suggestion](https://twitter.com/sebastienlorber/status/1537099054454296581):

1. purify on the server
2. using server components
3.  future: WIP HTML sanitizer API 

We already did `1`. We can try `2` when we migrate to React 18. And we hope that we can use `3` in the next few years.
For now, to add another layer of protection, I think `dompurify` should be a good solution.

PR #147, opened 2022-07-02
feat: write e2e tests for BestOfJs website

huyenuet:
## Feature
[x] Write e2e tests for BestOfJs website

### DIFF
--- a/e2e-testing/cypress.config.js
+++ b/e2e-testing/cypress.config.js
@@ -1,0 +1,9 @@
+const { defineConfig } = require("cypress");
+
+module.exports = defineConfig({
+  e2e: {
+    setupNodeEvents(on, config) {
+      // implement node event listeners here
+    },
+  },
+});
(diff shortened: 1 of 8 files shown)


michaelrambeau:
Hello @huyenuet 
Thank you very much for your work on this PR.
As I told our friend Hung, I'm not totally sure we need E2E testing with Cypress at this stage of the project as Hung has already invested a lot of time to add integration tests (with Jest Preview, of course!)

About the testing strategy itself, I'm not sure the test suite is robust if we target random CSS classes like `.css-wau20d` or even nested CSS selectors like `tfoot>tr a`. To me we should be able to change the HTML markup without breaking the tests.

huyenuet:
Hi @michaelrambeau, maybe it's not necessary for your project to have e2e testing at the current stage, so I'm going to close this PR. And thanks for your comment, I've changed to code to use test-id instead of CSS selectors.
I hope to have a chance to work with you in the future!

PR #152, opened 2022-08-07
Fix logos displayed as huge images in the README section of project details page

michaelrambeau:
## Goal

Fix a visual bug in some project README sections: some images are huge (taking all the available width, instead of a fixed width).

Examples:

- Expo project: https://bestofjs.org/projects/expo
- tRPC project: https://bestofjs.org/projects/trpc

Reason: Chakra UI comes with a Reset CSS file, whose style cause the problem.

https://github.com/chakra-ui/chakra-ui/blob/v1/packages/css-reset/src/css-reset.tsx

```css
img,
video {
  max-width: 100%;
  height: auto;
}
```

Rather than trying to override the styles (it was my first attempt but it's not trivial), I think it's cleaner to use our own Reset CSS file, created from Chakra UI's one.

## How to test

Open the projects mentioned above and check that the logo is displayed correctly (like in the original GitHub README!)

### Before

image

### After

image

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -4,6 +4,7 @@ import { Root } from "./root";
 import { initializeColorMode } from "components/core/color-mode";
 
 // Old-fashioned stylesheets
+import "./stylesheets/css-reset.css";
 import "./stylesheets/base.css";
 
 function start() {
--- a/src/root.tsx
+++ b/src/root.tsx
@@ -11,7 +11,7 @@ import { AuthProvider } from "containers/auth-container";
 
 export const Root = () => {
   return (
-    <ChakraProvider theme={customTheme} resetCSS={true}>
+    <ChakraProvider theme={customTheme} resetCSS={false}>
       <Router>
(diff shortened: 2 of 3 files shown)


nvh95 (approved):
Well. This is an interesting problem. The `height` property in `img` tag does not take effect. If the tRPC logo use `style="height: 130px;"`, there would be no problem.
tRPC logo looks fine if setting height by CSS

This PR LGTM. However, I'm not sure if it would break any other UI regarding `img` and `video`. An initiative is to use a visual testing tool in the future (e.g: https://www.chromatic.com/docs/test). I haven't used any of them but we can use one in the future for regression testing the UI.

michaelrambeau:
Thank you Hung for the good points!

Chakra UI components that include  `img` or `video` tags could be broken, that's the risk indeed.
Since ChakraUI has been introduced relatively recently, I'm confident but you are right, visual regression testing would provide more safety.

PR #177, opened 2023-08-05
fix: When no results are found, do not render useless headings

zawarudo:
# Fix 1:
## Goal
When I search for something that doesn't exist in projects and tags, I should not see the `Tags` and `Projects` headers.

## How to test

1. Open the search menu
2. Search for something that doesn't exist.
3. Confirm only `''No results found.''` renders in the results area.

## Screenshots

Error case:
image

Fixed:

image

image

### DIFF
--- a/apps/bestofjs-nextjs/src/components/search-palette/search-palette.tsx
+++ b/apps/bestofjs-nextjs/src/components/search-palette/search-palette.tsx
@@ -149,9 +149,9 @@ export function SearchPalette({ allProjects, allTags }: SearchProps) {
     });
   };
 
-  const isEmptyProjects = filteredProjects.length == 0,
-    isEmptyTags = filteredTags.length == 0,
-    isEmptySearchResults = isEmptyProjects && isEmptyTags;
+  const isEmptyProjects = filteredProjects.length == 0;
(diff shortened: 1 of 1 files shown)


apps/bestofjs-nextjs/src/components/search-palette/search-palette.tsx
```diff
@@ -149,6 +149,10 @@ export function SearchPalette({ allProjects, allTags }: SearchProps) {
     });
   };
 
+  const isEmptyProjects = filteredProjects.length == 0,
+    isEmptyTags = filteredTags.length == 0,
+    isEmptySearchResults = isEmptyProjects && isEmptyTags;
+
```
michaelrambeau, line 155: Even if it's standard JS, I would avoid the use of the `,` to separate variable declarations, could you use boring `const` for each line, to make things explicit and for consistency with the rest of the codebase?

zawarudo: Updated thanks.

michaelrambeau (approved):
👍 Thank you very much for entering this world, that's a great start!

PR #180, opened 2023-08-07
Adjust pagination limit for Next.js version

michaelrambeau:
## Goal

- Add more projects by page for `/projects` pages: from `10` to `30`, to match the current version. We could even go bigger (`50`?) as Next.js RSC are supposed to stream the HTML
- Fix missing "Star on GitHub" button for Best of JS, because of incorrect GitHub full name
- Add Vercel mention to the footer (because they provide a Pro plan to `bestofjs` org), we inline the SVG mark-up so that it works for both light and dark mode, using `fill="currentColor"`

## Screenshots

image


zawarudo (approved):
Looks solid 👍

Issue #105, opened 2021-05-28
Swap from google analytics to an open-source alternative


L1lith:
I was browsing the site and noticed that it loaded a script from google-analytics.com. Perhaps using an open-source and/or self-hosted solution would be better than google analytics from a privacy perspective. I haven't tried implementing anything like this myself, but perhaps [Matomo](https://matomo.org/) might be a good solution? This isn't a giant issue but I've been trying to reduce how much I expose myself and my users to companies that like to mine and sell data.

michaelrambeau:
Hello @L1lith 
Thank you for the idea.
I totally agree with you, actually I searched recently for some alternatives but I couldn't find anything that was free.

The current number of monthly Page Views is around 90,000 according to Google Analytics but I guess the real numbers are bigger since developers often use solutions to avoid being tracked.

So I'm going to investigate seriously about this issue and try to implement something.
Just in case, have you any feedback about https://www.cloudflare.com/analytics/ ?

L1lith:
@michaelrambeau Here's one [blog](https://www.ctrl.blog/entry/review-cloudflare-analytics.html) I found detailing the merit of Cloudflare analytics. It appears to be overall a good option, though it might be worth looking into the cookie issue talked about inside the article. 

Overall finding quality free services is a difficult thing. Free software is fairly abundant but it all requires a back-end to host it (which isn't free). I think for now Cloudflare would be the best choice.

michaelrambeau:
Thank you @L1lith  for the feedback, Cloudflare seems simplest solution to integrate as it's just a snippet and no setup is required, so I'm going to try it ASAP.

We have a tag "analytics" in _Best of JS_ actuallly: https://bestofjs.org/projects?tags=analytics

What do you think of the solution Ackee? https://docs.ackee.electerious.com/

I like the fact it runs on Node.js and on MongoDB (the stack I'm familiar with), it could be installed easily on Vercel, it seems.

michaelrambeau:
## Before saying goodbye

Before removing GoogleAnalytics, for the records, here is a screenshot about the number of monthly Page Views since the creation of the project.
There was a peak of 166k PageViews in June 2019.
The number is steadily decreasing since that: 94k in April 2021.

![image](https://user-images.githubusercontent.com/5546996/119984279-2e09f380-bffc-11eb-94b1-bb081a98ea70.png)

[...]

PR #181, opened 2023-08-07
Try Vercel analytics for Next.js app

michaelrambeau:
## Goal

It would be nice to get data about the traffic and the pages visited by users, without compromising user's privacy and without introducing heavy scripts in the client.

Now that we have more server-centric application, it should run on the server, ideally.

Let's try Vercel Analytics: https://vercel.com/docs/concepts/analytics/quickstart

Unfortunately it's not a server-side only solution:

> If everything is set up properly, you should be able to see a Fetch/XHR request in your browser's Network tab from /_vercel/insights/view when you visit any page

Related: #105

### DIFF
--- a/apps/bestofjs-nextjs/package.json
+++ b/apps/bestofjs-nextjs/package.json
@@ -30,6 +30,7 @@
     "@types/fs-extra": "^11.0.1",
     "@types/lodash": "^4.14.195",
     "@types/numeral": "^2.0.2",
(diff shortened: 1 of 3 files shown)


michaelrambeau:
## Screenshot about data for the beta version

24 hours later, the good news is that we have users who tried the beta version, from the link that was added to the home page of the app.

image

PR #185, opened 2023-08-12
Handle `isFeatured` flag to select "Featured projects"

michaelrambeau:
## Goal

In the backend, a status option "featured" was added to be able to detect the projects considered as "Featured":
 https://github.com/bestofjs/bestofjs-backend/commit/21ed1533e9048d3284d1a57dffcfbe9f96684a9d
 
Before we were only relying on the existence of a specific icon for the project, which was not great as we want the ability to setup a custom icon without making the project "Featured".

Note: there is no specific criteria to make a project "Featured" we only want highlight a subset of important projects.
There are currently 237 projects showing up, I'm cleaning up up to try to make this  number close to 200.

Related PR for the beta version: https://github.com/bestofjs/bestofjs/pull/184

### Before

image

### After

image

### DIFF
--- a/apps/bestofjs-webui/src/global.d.ts
+++ b/apps/bestofjs-webui/src/global.d.ts
@@ -22,6 +22,7 @@ declare namespace BestOfJS {
     npm: string;
     downloads: number;
     icon: string;
+    isFeatured: boolean;
   }
 
   // Project handled in the state container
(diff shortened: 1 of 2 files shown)


zawarudo (approved):
Makes sense, everything checks out 👍

PR #191, opened 2023-08-23
feat: improve hover ux of search dialog

zawarudo:
## Goal
On search results add cursor pointer to hover-able elements,

## How to test
Hover over all clickable elements in a search dialog.

## Screenshots

image
image
image
image

### DIFF
--- a/apps/bestofjs-nextjs/src/components/search-palette/search-palette.tsx
+++ b/apps/bestofjs-nextjs/src/components/search-palette/search-palette.tsx
@@ -183,7 +183,7 @@ export function SearchPalette({ allProjects, allTags }: SearchProps) {
             {false && <LoadingProject project={allProjects[1000]} />}
             {false && <LoadingTag tags={[allTags[10]]} />}
             {currentTags.length > 0 && (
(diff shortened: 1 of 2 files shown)


michaelrambeau:
Thank you for caring about the UI/UX of the search feature, that is the key feature of _Best of JS_!

I may be wrong (or we may need the opinion of a third person) but I find the orange color that appears only when rows are selected distracting.
it's a table and the whole row is clickable, so I don't think we need this effect.
To confirm this impression, I checked sites with similar keyboard-friendly search palette in dark mode:

#### [TailwindCSS](https://tailwindcss.com/)

image

#### [Next.js](https://nextjs.org/)

image

#### [Radix](https://www.radix-ui.com/primitives/docs/overview/introduction)

image


#### [NextUI](https://nextui.org/)

image

#### [Adobe Spectrum](https://react-spectrum.adobe.com/)

One of the reasons is that we might need to highlight the matches, like React Spectrum does

image

michaelrambeau (approved):
Thank you, it looks good on light and dark modes 👍

PR #192, opened 2023-08-26
Generate dynamic OG images

michaelrambeau:
## Goal

We need OG images (Open Graph) to make _Best of JS_ links more attractive, when links are shared on social networks.

OG images can be generally dynamically with this very cool feature: https://vercel.com/docs/functions/edge-functions/og-image-generation

Under the hood it uses a package called Satori: https://og-playground.vercel.app/

We could use this feature to generate images that display data, depending on the page requested.

- The default could image a summary of the daily, for example: `Hottest Project Today: #1 VSCode +200 stars`
- For project details page, we could show a summary of the project (TBD)

This feature require the Edge run time, so we can't use the filesystem to fetch the JSON file we use as a temporary cache to avoid pinging too many times the static API JSON file... let's check that it has no impact on performance.

## How to test

The image is generated when pinging the URL http://localhost:3000/api/og

Share the link on Slack or Discord to visualize the image.

## Screenshots

First iteration

image


michaelrambeau:
## Iteration 2

### Default

http://localhost:3000/api/og

image

### Project details page

http://localhost:3000/api/og/projects/autoanimate

image

michaelrambeau:
## Issue with the local build

Fetching data from the remote JSON file instead of the local one does not seem to cause problems of performance, the only issue I had is about my build in local, as I have a frequent error "terminated".

My asumption is that it's because of the number of fetch requests but I'm not sure.
It has to do with the latency as eveyrthing is fine on Vercel, no single deploy about the PR failed.

image

Switching from Nod.js 18 to Node.js 20 seems to fix the error.

michaelrambeau:
## Bonus

You can preview the OG images right from the Vercel dashboard!

image

apps/bestofjs-nextjs/src/app/backend.ts
```diff
@@ -352,14 +346,8 @@ async function fetchProjectData(): Promise<{
...
   console.log(`Fetching JSON data from ${url}`);
-  const options = { next: { revalidate: 60 * 60 } }; // Revalidate in one hour
-  return fetch(url, options).then((res) => res.json());
-}
-
-function fetchDataFromFileSystem() {
-  console.log("Fetch from the file system");
-  const filepath = path.join(process.cwd(), "public", "data/projects.json");
-  return fs.readJSON(filepath);
+  // const options = { next: { revalidate: 60 * 60 } }; // Revalidate in one hour
```
zawarudo, line 349: Why did we remove revalidate here and
[here](
https://github.com/bestofjs/bestofjs/pull/192/files#diff-d55bd9b8687cc88e8524c3ad2be17f7251d45491f5ebc061d2d227eef3a7899aL36)?

zawarudo (approved):
Looks good!

Thanks for letting me have a look.

Just have one comment about revalidation I want to clarify.

apps/bestofjs-nextjs/src/app/backend.ts
michaelrambeau: The comment said:

> "TODO remove when page rebuilt is triggered automatically"

It's exactly what we are doing now, the site is rebuilt everyday on Vercel, just after the static API (the JSON file) has been rebuilt, so we don't need to tell Next.js to revalidate the page. The page is correct until the site is rebuilt, every 24 hours.

PR #199, opened 2023-09-03
Improve light mode, cleanup component before releasing the beta version

michaelrambeau:
## Goal

Cleanup UI component and improve UI details about the cards in the project details page.

- add an icon for external links to GitHub and NPM sites
- Remove the dash effect used to separate section of the cards
- remove the Card component inside `components/core` as we have another one in `components/ui` (from `shadcn/ui` project that we use to "bootstrap" the Next.js version)
 
Also: 

- don't show the exact time of the last update, only the relative time (because the page is rendered server-side, it was not the local time that was displayed)
- Reset CSS values for the light version, using values from the "stone" theme, to make colors less blueish and closer to the current version

https://ui.shadcn.com/themes

## How to test

Check `/projects/:id` URLs

## Screenshots

image

image


apps/bestofjs-nextjs/src/app/page.tsx
```diff
@@ -265,10 +265,8 @@ function MoreProjectsSection({
           .
         </p>
         <p>
-          Data is updated from GitHub everyday, the last update was{" "}
-          <b>{fromNow(lastUpdateDate)}</b> (at{" "}
-          {timeOnlyFormat.format(lastUpdateDate)}
-          ).
+          Data is updated from GitHub every 24 hours, the last update was
+          <b>{fromNow(lastUpdateDate)}</b>.
```
zawarudo, line 269: PR looks great! Just needs a space here between `the last update` and the `fromNow` call.

zawarudo (approved):

PR #200, opened 2023-09-03
Split backend layer into multiple services

michaelrambeau:
## Goal

Refactoring to split and rename the the `searchClient` object used to communicate with the server.

Now the API is made of 3 services:

- `projects`
- `tags`
- `services`

From the components, instead of calling `searchClient.findProjects`, we call `api.projects.findProjects`.

Reason: previous the `backend` was way too big, making difficult to understand what is going on in the server side 😅 

This PR also adds a placeholder page for the URL `/timeline`, just as a reminder that I have to port this page, sorry for the extra noise!

#### API layer

```tsx
export function createAPI() {
  async function getData() {
   // fetching logic 
  }

  const context = {
    getData,
  };

  const projectsAPI = createProjectsAPI(context);
  const tagsAPI = createTagsAPI(context);
  const hallOfFameAPI = createHallOfFameAPI(context);

  return {
    projects: projectsAPI,
    tags: tagsAPI,
    hallOfFame: hallOfFameAPI,
  };
}
```

#### How it's called from the async components

```tsx
const {projects} = await api.projects.findProjects({criteria: {}});

const {tags} = await api.tags.findTags({criteria: {}});
```


apps/bestofjs-nextjs/src/server/api-hall-of-fame.tsx
```diff
@@ -0,0 +1,44 @@
+import { Data, FETCH_ALL_PROJECTS_URL } from "./api-utils";
+
+export function createHallOfFameAPI({
+  getData,
+}: {
+  getData: () => Promise<Data>;
+}) {
+  return {
+    async findMembers() {
+      const { populate, projectsBySlug } = await getData();
```
zawarudo, line 10: `getData` can be undefined here right?

apps/bestofjs-nextjs/src/server/api-projects.tsx
```diff
@@ -0,0 +1,148 @@
+import * as mingo from "mingo";
+import { RawObject } from "mingo/types";
+
+import { filterProjectsByQuery } from "@/components/search-palette/find-projects";
```
zawarudo, line 4: I'm looking at `find-projects.ts`, I'm a little surprised this comes from `@/components`

It seems like a util function mostly used by `search-palette`

Maybe it would be more clear to me if the file structure was changed.

`components/search-palette/utils/find-projects`

michaelrambeau: Good catch! We need a kind of `shared` folder for functions used by both components and the API layer, as the server should not import functions coming from components.

apps/bestofjs-nextjs/src/server/api-hall-of-fame.tsx
michaelrambeau: `getData` is part of the context that each of the 3 APIs receive, check the types in VSCode:

image

Maybe I should make it more explicit by creating a specific type `Context`.

apps/bestofjs-nextjs/src/server/api.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const data = await fetchDataFromRemoteJSON();
+    return data as RawData;
+  } catch (error) {
+    console.error("Unable to fetch data!", (error as Error).message);
+    throw error;
+  }
+}
+
+function fetchDataFromRemoteJSON() {
+  const url = FETCH_ALL_PROJECTS_URL + `/projects.json`;
```
zawarudo, line 72: This is where we get all of the alerts from when building right?

Can't we cache this call after running it once?

Or better yet do it in `fetchProjectData()`

zawarudo (approved):

apps/bestofjs-nextjs/src/server/api-hall-of-fame.tsx
michaelrambeau: Done here: https://github.com/bestofjs/bestofjs/pull/200/commits/8d7b57b050efd060720d3ead67f00d5e72464d95

apps/bestofjs-nextjs/src/server/api-projects.tsx
michaelrambeau: Done here: https://github.com/bestofjs/bestofjs/pull/200/commits/4e3d1ba28ef9299202186586631729e6c96f376a

apps/bestofjs-nextjs/src/server/api.tsx
michaelrambeau: Caching is tricky as I realized the previous solutions caused issues in the real site.

The best would be to be able to reproduce in local what Vercel does in the live site with `fetch` calls 🤔

zawarudo (approved):
I like it 👍

PR #204, opened 2023-09-10
Fix local build using a local JSON file for data

michaelrambeau:
## Goal

Fix local build and re-introduce the local JSON file used as a local cache and re-built every day.

Because we have edge functions that generate images dynamically and that can access the file system, there are 2 versions of the API layer:

- one that uses the local JSON data `import {api} from "@app/server/api"` (the default one used for all pages) 
- one that uses that fetches JSON data remotely `import {api} from "@app/server/api-remote-json"`, used by API end-points with the `edge` runtime
 
The gist: the function that creates the `api` object used to query data now accepts a `fetchProjectData` function.
The strategy to fetch data (using local or remote JSOB) is considered as a dependency.

```ts
function createAPI(fetchProjectData: () => Promise<RawData>) {
  const { projects, tags, date } = await fetchProjectData();
  // ...
}
```

This way we can create different `api` objects that use a different strategy.

## How to test

Build locally end ensure there is no more errors:

```sh
pnpm -F bestofjs-nextjs build
```

## Thinking about the future

The current strategy of loading a big JSON file is not great, the implementation could be simplified if we had a real API server that responds to queries, instead of making the Next.js app handle the JSON data.

### DIFF
--- a/apps/bestofjs-nextjs/.eslintrc.cjs
+++ b/apps/bestofjs-nextjs/.eslintrc.cjs
@@ -1,3 +1,5 @@
+const path = require("path");
+
 module.exports = {
   $schema: "https://json.schemastore.org/eslintrc",
   root: true,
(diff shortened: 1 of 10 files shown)


zawarudo (approved):
Works on my end.

Clean separation between local / remote API.

Everything makes sense.

PR #211, opened 2023-10-13
Simplify NPM dependencies layout

michaelrambeau:
## Goal

Fix the layout of the dependencies table, removing table headers "Dependencies on Best of JS" and "Dependencies not on Best of JS".

The layout is currently broken on production:

image

I think it looks better without header at all? (screenshot below)

## How to test

Check projects that have a related NPM package, with different scenarios.

Only deps that are NOT on Best of JS:

http://localhost:3000/projects/axios

A mix of projects on and not on Best of JS: 

- http://localhost:3000/projects/tldraw
- http://localhost:3000/projects/genql

## Screenshots

image

### DIFF
--- a/apps/bestofjs-nextjs/src/app/projects/[slug]/project-details-npm/dependencies-section.tsx
+++ b/apps/bestofjs-nextjs/src/app/projects/[slug]/project-details-npm/dependencies-section.tsx
@@ -1,5 +1,5 @@
 import { Badge } from "@/components/ui/badge";
-import { Card, CardHeader } from "@/components/ui/card";
+import { Card } from "@/components/ui/card";
 import {
   Collapsible,
   CollapsibleContent,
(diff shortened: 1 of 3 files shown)


apps/bestofjs-nextjs/src/app/projects/[slug]/project-details-npm/dependencies-section.tsx
```diff
@@ -39,26 +39,11 @@ export async function DependenciesSection({
       <CollapsibleContent className="space-y-4 py-4">
         {projects.length > 0 && (
           <Card>
-            <CardHeader>
```
Comment (zawarudo) [inline apps/bestofjs-nextjs/src/app/projects/[slug]/project-details-npm/dependencies-section.tsx:42]: I think we can remove the `CardHeader` from the imports now.

zawarudo (changes requested):

apps/bestofjs-nextjs/src/app/projects/[slug]/project-details-npm/dependencies-section.tsx
Comment (michaelrambeau) [reply apps/bestofjs-nextjs/src/app/projects/[slug]/project-details-npm/dependencies-section.tsx:42]: Good catch, I will set up the ESLint rule that should not allow unused variables in a separate PR.

PR #212, opened 2023-10-13
Lint/no unused imports

zawarudo:
## Goal
Linter should fix / warn about unused variables.


michaelrambeau (approved):
Thank you for the PR, these unused imports not caught by the linter were very annoying.

Soon we could try to introduce a standard setup using something like `extends": "eslint:recommended"` to catch the most common problems.

PR #214, opened 2023-10-14
ESLint recommended setup, cleanup of errors and warnings

michaelrambeau:
## Goal

- Add TypeScript ESLint recommended setup (to catch basic problems such as unused variables)
- Fix ESLint errors and warnings
- Remove `eslint-plugin-unused-imports` plugin as the feature is already included by the TypeScript ESLint recommended set
- Remove warning when running commands from the PNPM root level

```
The field "pnpm" was found in /Users/michaelrambeau/dev/bestofjs/bestofjs-monorepo/apps/bestofjs-webui/package.json. This will not take effect. You should configure "pnpm" at the root of the workspace instead
```

This PR also adds a basic workflow on GitHub actions to ensure `test` and `lint` command pass, before a PR is merged.

## How to test

- Run `pnpm install` at the root level
- check the absence of warnings
- Run `pnpm -F bestofjs-nextjs lint` there should be no errors and no warnings

## Screenshots

Warning removed:

image

Lint command:

image


michaelrambeau:
## What is missing

How do we ensure lint command passes before a PR is merged?

- we could setup a Git hook in local after before every push but the command is quite slow, so it's a bit annoying
- the best could be a "check" visible in the PR, that prevents the PR from being merged if the lint command (and the tests too) don't pass

apps/bestofjs-nextjs/src/app/projects/[slug]/get-project-details.tsx
```diff
@@ -11,14 +23,18 @@ async function fetchProjectDetailsData(fullName: string) {
...
-function mergeProjectData(project: BestOfJS.Project, details: any) {
+function mergeProjectData(
+  project: BestOfJS.Project,
+  details: ProjectDetailsRawData
+) {
   const {
     npm,
     bundle,
-    packageSize,
+    // packageSize,
```
Comment (zawarudo) [inline apps/bestofjs-nextjs/src/app/projects/[slug]/get-project-details.tsx:37]: Saving these for later? ;)

apps/bestofjs-nextjs/src/components/search-palette/search-palette.tsx
```diff
@@ -123,7 +124,8 @@ export function SearchPalette({ allProjects, allTags }: SearchProps) {
   const onSelectProject = (itemValue: string) => {
     const projectSlug = itemValue.slice("project/".length);
     const project = allProjects.find((project) => project.slug === projectSlug);
-    setSelectedItem({ type: "project", value: project! });
+    invariant(project);
```
zawarudo, line 127: Not including an error message for invariant to throw?

zawarudo (changes requested):

apps/bestofjs-nextjs/src/components/search-palette/search-palette.tsx
michaelrambeau: It's not required as I use `tiny-invariant` for its "type narrowing" feature, to avoid non null assertions (rather than for run-time checks) but I can add a message, OK, let's be safe :)

https://github.com/alexreardon/tiny-invariant

michaelrambeau:
We have now a basic pipeline to ensure `test` and `lint` commands pass, look at this! 😄 

The only downside: it adds 35 seconds before the app is built (and of course it will grow over time with the codebase!)

image

zawarudo (approved):

Issue #210, opened 2023-10-09
Add a specific OG image and description for project list pages


michaelrambeau:
Thanks to the dynamic generation of images provided by [Satori project](https://og-playground.vercel.app/), we provide a specific OG (Open Graph) image for the 2 following pages:

- the top page: `/` path
- the project's detail page `/projects/:slug` path
 
It would be great to get a specific OG image for project list pages (`/projects` path, with or without query string), showing an overview of the 3 first projects of the list.

Right now, it's the top page OG image that is displayed, because metadata is inherited from the parent page and `/app/projects/page.tsx` does not specify any

Docs about metadata: https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image

 ### Screenshots

Top page

Image URL: https://bestofjs.org/api/og

<img width="629" alt="image" src="https://github.com/bestofjs/bestofjs/assets/5546996/f29cf81a-2fa9-411f-8cf6-bf7b76ed0000">

For React project

Image URL: https://bestofjs.org/api/og/projects/react

<img width="421" alt="image" src="https://github.com/bestofjs/bestofjs/assets/5546996/765be5d8-9d01-47f3-ad15-72fb90b38433">

### To be defined

What to show as the caption, before the 3 rows?
Something similar to what we show in the `/projects` page URLs?

There are different scenarios:

- no query string at all: it's the list of all available projects
- one or several tags
- without or without `query` parameter when showing search results
 
For example, for `Vue.js` projects, the tag icon, followed by the name of the tag:

![image](https://github.com/bestofjs/bestofjs/assets/5546996/7b5202cb-1374-4b48-bb66-a00342dd1718)

Evomatic:
Hello @michaelrambeau 

After taking a look, I want to make sure I have a correct understanding of what we want to do here!

So we want to add a new OG image that is specific for the path `/projects`, because right now, linking `https://bestofjs.org/projects` gives us the same OG image used at the top page. 

Things to consider for this new specific OG image:

- If no query string, list all available projects. Or do we want to list the first 3 (depending on sort method), similar to the top page list?
- If tags are included (one to three tags) how should we show that in the caption?
- If `query` parameter is applicable, how should we show this?

 As you mentioned I think doing something similar to what is shown for `/projects` is good. For example, if using tags then show tag icon, followed by name of tag. Then show the list projects with the tag below.

If no tags selected then show "All Projects" followed by number of projects

[...]

PR #215, opened 2023-10-16
WIP: Add specific OG image for project list pages

Evomatic:
## Goal
Add a specific OG image for project list pages `/projects`.

More details about this PR can be found in the issue: https://github.com/bestofjs/bestofjs/issues/210
## How to test

- Run `pnpm install` at root of repository
- Pull this branch and checkout locally
- Go to [http://localhost:3000/api/og/projects](http://localhost:3000/api/og/projects) - there should be a new OG image that is similar to what is seen at [http://localhost:3000/projects](http://localhost:3000/projects)
- try adding `query` and `tags` to the end of `projects/`. For example: `?query=tool&tags=cli` - The OG image should display the first 3 projects based on the search params.


apps/bestofjs-nextjs/src/app/api/og/route.tsx
```diff
@@ -35,7 +35,7 @@ export async function GET() {
   );
 }
 
-function ProjectRow({
+export function ProjectRow({
   project,
   rank,
 }: {
```
michaelrambeau, line 41: Thank you for the PR!
You should move this function to a shared file as I think you cannot use named `export` for special files such as `page.tsx` or `route.tsx`.

I think it's the reason why the deploy failed on Vercel.

```
[16:27:04.943] Failed to compile.
[16:27:04.943] 
[16:27:04.943] src/app/api/og/route.tsx
[16:27:04.944] Type error: Route "src/app/api/og/route.tsx" does not match the required types of a Next.js Route.
[16:27:04.944]   "ProjectRow" is not a valid Route export field.
[16:27:04.944] 
[16:27:05.052]  ELIFECYCLE  Command failed with exit code 1.
[16:27:05.061] Error: Command "pnpm build" exited with 1
```

apps/bestofjs-nextjs/src/app/api/og/projects/route.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Box, generateImageResponse, mutedColor } from "@/app/api/og/og-utils";
+
+import { ImageLayout } from "../og-image-layout";
+import { ProjectRow } from "../route";
+
+export const runtime = "edge";
+
+export async function GET(_req: Request) {
+  const { tags, query } = useSearchParams(_req.url);
+  const { projects } = await api.projects.findProjects({ query });
```
michaelrambeau, line 15: I think you are missing some parameters when calling `api.projects.findProjects()`.
The `query` is about the text search but we have also the "tags", the sort filter, the pagination...

Basically we want to perform the same search as in the regular `/projects` page, using all information provided by the `ProjectSearchQuery` object returned by `parseSearchParams` function.

```ts
  const { tags, sort, page, limit, query } = parseSearchParams(searchParams);
  const sortOption = getSortOption(sort);

  const { projects } = await api.projects.findProjects({
    criteria: tags.length > 0 ? { tags: { $all: tags } } : {},
    query,
    sort: sortOption.sort,
    skip: limit * (page - 1),
    limit,
  });
```

That also means we could do some refactoring to extract the logic shared by the page and the image, since they both need to fetch an array of projects given search parameters, but we can do this later.

Evomatic: > I think you are missing some parameters when calling `api.projects.findProjects()`. The `query` is about the text search but we have also the "tags", the sort filter, the pagination...
> 
> Basically we want to perform the same search as in the regular `/projects` page, using all information provided by the `ProjectSearchQuery` object returned by `parseSearchParams` function.
> 
> ```ts
>   const { tags, sort, page, limit, query } = parseSearchParams(searchParams);
>   const sortOption = getSortOption(sort);
> 
>   const { projects } = await api.projects.findProjects({
>     criteria: tags.length > 0 ? { tags: { $all: tags } } : {},
>     query,
>     sort: sortOption.sort,
>     skip: limit * (page - 1),
>     limit,
>   });
> ```
> 
> That also means we could do some refactoring to extract the logic shared by the page and the image, since they both need to fetch an array of projects given search parameters. Basically the snippet above could be inside a shared function), but we can do this later.

I am bit confused as to what is happening here:

```
 const { projects } = await api.projects.findProjects({
    criteria: tags.length > 0 ? { tags: { $all: tags } } : {},
    query,
    sort: sortOption.sort,
    skip: limit * (page - 1),
    limit,
  });
```
Does this object represent the projects filtered by params? 

I agree in extracting the shared logic since a lot of the code I am writing is shared in the page.  Could be good to tackle next!

michaelrambeau: Sorry for the confusing naming, let me describe the parameters of the `findProjects` method:

- `criteria` is an object that expresses the search filter. It uses MongoDB syntax, `$all` being an operator to include elements that all exist inside an array
- `query`: is the text query when users enter some text (I think I should rename it `textQuery` ?)
- `skip` and `limit` is about pagination, I think you are familiar with that if have done SQL requests for example
- `sort` describes the sort to be applied, again using the MongoDB syntax. E.g. `{ stars: -1 }` to sort by stars in the descending order

Evomatic:
Hi @michaelrambeau - I have this pretty much in a working state but there could possibly be some improvements

Tags:
image

Query:
image

All:
image

As you mentioned before, there could be some refactoring to get rid of redundant code.

- This snippet you mentioned above can be moved to a shared method
- `getSortOption` method could also be moved and become a shared method


I can look into refactoring, but would we want to include it in this PR or have it be a separate PR in order to keep this one from becoming too big?

apps/bestofjs-nextjs/src/app/api/og/projects/route.tsx
```diff
@@ -0,0 +1,96 @@
...
+} from "@/components/project-list/sort-order-options";
+import { api } from "@/server/api-remote-json";
+import { Box, generateImageResponse, mutedColor } from "@/app/api/og/og-utils";
+
+import { ImageLayout } from "../og-image-layout";
+import { ProjectRow } from "../og-utils";
+
+export const runtime = "edge";
+
+export async function GET(_req: Request) {
```
michaelrambeau, line 20: It's better to avoid the prefix `_` as it sometimes conveys a special meaning.

Example: https://nextjs.org/docs/app/building-your-application/configuring/draft-mode#step-1-create-and-access-the-route-handler

apps/bestofjs-nextjs/src/app/api/og/projects/route.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const { searchParams } = new URL(url);
+  const projectSearchParams: ProjectPageSearchParams = {
+    tags: searchParams.getAll("tags") || undefined,
+    query: searchParams.get("query") || undefined,
+    page: searchParams.get("page") || undefined,
+    sort: searchParams.get("sort") || undefined,
+    limit: searchParams.get("limit") || undefined,
+  };
+  return parseSearchParams(projectSearchParams);
+}
```
michaelrambeau, line 77: In the React world, `useXXX` is a convention to name a hook (that usually combines other hooks like `useState` or `useContext`).

Here we are server-side, in an API route, it's not a React hook, so I would name this function differently.

michaelrambeau:
From you screenshots, it seems we are fetching more than the 3 projects we need (I can see the beginning of a 4th project 😄 ), have you set the `limit` parameter correctly?

michaelrambeau:
> Hi @michaelrambeau - I have this pretty much in a working state but there could possibly be some improvements
> 
> Tags: image
> 
> Query: image
> 
> All: image
> 
> As you mentioned before, there could be some refactoring to get rid of redundant code.
> 
>     * This snippet you mentioned above can be moved to a shared method
> 
>     * `getSortOption` method could also be moved and become a shared method
> 
> 
> I can look into refactoring, but would we want to include it in this PR or have it be a separate PR in order to keep this one from becoming too big?

We can make the refactoring after this PR is merged, to keep it focused on the feature, indeed.

michaelrambeau:
About the content of the images:

- by default, I think it should display the total number of stars (otherwise we don't know the meaning of the `+`), like in the regular pages, since projects are sorted by number of stars, unless a specific sort order is applied.
- if a sort order is applied, we could display the related number ("+400") but that would mean a specific caption that says "by stars added today" for example.

apps/bestofjs-nextjs/src/app/api/og/projects/route.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+export async function GET(_req: Request) {
+  const { tags, query, sort, page, limit } = useSearchParams(_req.url);
+  const sortOption = getSortOption(sort);
+  const { projects, total } = await api.projects.findProjects({
+    criteria: tags.length > 0 ? { tags: { $all: tags } } : {},
+    query,
+    sort: sortOption.sort,
+    skip: limit * (page - 1),
+    limit,
```
michaelrambeau, line 28: The limit should be set to `3` and we should ignore the `limit` from the query string.

Evomatic:
> About the content of the images:
> 
>     * by default, I think it should display the total number of stars (otherwise we don't know the meaning of the `+`), like in the regular pages, since projects are sorted by number of stars, unless a specific sort order is applied.
> 
>     * if a sort order is applied, we could display the related number ("+400") but that would mean a specific caption that says "by stars added today" for example.

Ok, so the number we see now (in the images above) are the number of stars added today and not the total number? And the starts added today are a result of the sorting option "By stars added..."?

michaelrambeau:
The OG image associated to a given page is supposed to reflect the content of the page, in a compact way.

The home page showing the hottest project of the day, that is to say the projects sorted by numbers of stars added today, the OG image includes 3 projects with the related numbers (the "TOP 3" of the day), this is why we show something like `+ 100`.

However, in our case, we have to deal with all types of filter, not just "by number of stars added today", it can any depending on the user's choice.

To keep things simple,  I suggest to make a simplified version of the table:

- do not show the ranking numbers (`#1`) in the first column because it's not really a ranking
- instead of showing the stars added since yesterday, show the total number of stars.

It also means it may be better to not re-use the existing table of projects (reverting some changes made before, sorry!)

What do you think, does it make sense?

apps/bestofjs-nextjs/src/app/api/og/projects/route.tsx
```diff
@@ -0,0 +1,147 @@
...
+    criteria: tags.length > 0 ? { tags: { $all: tags } } : {},
+    query,
+    sort: sortOption.sort,
+    skip: NUMBER_OF_PROJECTS * (page - 1),
+    limit: NUMBER_OF_PROJECTS,
+  });
+
+  return generateImageResponse(
+    <ImageLayout>
+      {ImageCaption({ tags, query, sortOption })}
```
michaelrambeau, line 39: Can you use the JSX syntax `<ImageCaption>` instead of opening a block `{}`?

michaelrambeau (approved):
Thank you so much Evan for your first PR in the _Best of JS_ world, great job! 🎉 

It was not as easy as expected because of the missing requirements (sorry!) and the fact that these OG images are also a piece of UI, they deserve some attention.

I have checked using the Preview URL generated by Vercel and it works well!

Screenshot from Discord when including the URL about the "UI Framework" projects, with a specific sort:

image

Evomatic:
Thank you! It was my pleasure to contribute, I definitely learned a lot!

PR #219, opened 2023-10-16
e2e: add playwright as an optional dependency

zawarudo:
## Goal
Add playwright e2e as an optional dependency that the user has to opt-in to setup and use.

Note:

- Do we want to include this in the monorepo root or its fine as a `next-js` version only dependency?
- For the added e2e commands, we might want to rename them.
- Included tests run are from `example.spec.ts`  which do some simple tests at https://playwright.dev/, no bestofjs tests are included yet.
- By default tests will be run on all 3 browsers, chromium, firefox and webkit using 3 workers. This can be configured in the [playwright.config file](https://playwright.dev/docs/test-configuration).

## How to test

```
cd ./apps/bestofjs-nextjs
```

```
# install deps
pnpm i
```

```
# running e2e tests should fail because e2e is optional
pnpm test:e2e
```

```
# install playwright
pnpm e2e:init
```

```
# tests should run and pass
pnpm test:e2e
```

## Screenshots

Running e2e without installing
image

Running e2e tests after installing

Note: One test is failing due to a timeout,
These tests are bundled with playwright and should pass
this is why I wanted to discuss playwright as my local tests are flaky because of this.
I'm not sure if my CPU is too weak and its causing issues...

image

### DIFF
--- a/apps/bestofjs-nextjs/.gitignore
+++ b/apps/bestofjs-nextjs/.gitignore
@@ -37,3 +37,6 @@ yarn-error.log*
 
 # Data generated at build time
 public/data
+/test-results/
+/playwright-report/
+/playwright/.cache/
--- a/apps/bestofjs-nextjs/e2e/example.spec.ts
+++ b/apps/bestofjs-nextjs/e2e/example.spec.ts
@@ -1,0 +1,18 @@
+import { test, expect } from '@playwright/test';
+
+test('has title', async ({ page }) => {
+  await page.goto('https://playwright.dev/');
+
+  // Expect a title "to contain" a substring.
+  await expect(page).toHaveTitle(/Playwright/);
+});
+
+test('get started link', async ({ page }) => {
+  await page.goto('https://playwright.dev/');
+
+  // Click the get started link.
+  await page.getByRole('link', { name: 'Get started' }).click();
+
+  // Expects page to have a heading with the name of Installation.
+  await expect(page.getByRole('heading', { name: 'Installation' })).toBeVisible();
+});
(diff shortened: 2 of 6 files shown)


michaelrambeau:
Thank you @zawarudo for your research on this topic!

I didn't know about optional dependencies, that is an interesting approach.

I wonder if doing this really installs the version specified in package.json:

```sh
pnpm exec playwright install
```

maybe we should do this instead?

```sh
pnpm install --no-optional
```

to ensure we read the package.json file to install `playwright` dependency?

Also, in our test, we may need an initial step that checks if playwright is installed and throw a readable error "You should install doing this and that" maybe ?

Another approach to avoid installing the E2E dependencies when running `pnpm i` at the root level would be to move the `E2E` out of the workspace, considering it as another application that has to be installed specifically.

```
.
├── apps
│   ├── bestofjs-nextjs
│   └── bestofjs-webui
├── e2e
```

zawarudo:
> I didn't know about optional dependencies, that is an interesting approach.
> 
> I wonder if doing this really installs the version specified in package.json:
> 
> ```shell
> pnpm exec playwright install
> ```

I looked in my `node_modules/` and we do install the correct version.

> Also, in our test, we may need an initial step that checks if playwright is installed and throw a readable error "You should install doing this and that" maybe ?
> 
> Another approach to avoid installing the E2E dependencies when running `pnpm i` at the root level would be to move the `E2E` out of the workspace, considering it as another application that has to be installed specifically.
> 
> ```
> .
> ├── apps
> │   ├── bestofjs-nextjs
> │   └── bestofjs-webui
> ├── e2e
> ```
> 
> I wonder if it would be simpler?


I think It would be good to make some real tests that pass first, then I will try moving the folder.

zawarudo:
I'll remake this PR since adding it as optional dependency isn't a well documented use case.
(seems like a premature optimization)

PR #226, opened 2023-10-21
Set up env variables for API root URLs with T3 Env

michaelrambeau:
## Goal

Set up API root URLs using `T3 Env` package instead of hard-coding URLs in several places.

Bonus:

- Validation of the env vars at build time
- Autocompletion in your IDE!

This approach will be useful when we implement server-side features in the app that require an access to the database, for example.

Doc: https://env.t3.gg/docs/nextjs

## Screenshots

Autocomplete in action!

image


zawarudo (approved):
LGTM

PR #227, opened 2023-10-22
Fix 404 error in the console because of redirect

michaelrambeau:
## Goal

Fix a 404 error showing up in the console, about React Server Components for `/rankings/monthly` path:

```        
GET https://bestofjs.org/rankings/monthly?_rsc=acgkz 404 (Not Found)
```

This path, used for the "Monthly" link in the menu, redirects the user to the latest month rankings page (E.g. `/rankings/monthly/2023/9`).

The redirect was setup using a API route (`route.tsx`).
Now it's setup using the main Next.js config file (which is a `.mjs` file, not TypeScript).

## How to test

- Ensure the error is gone
- Ensure the redirect works as expected: click on "Monthly" in the top bar

### DIFF
--- a/apps/bestofjs-nextjs/next.config.mjs
+++ b/apps/bestofjs-nextjs/next.config.mjs
@@ -1,4 +1,4 @@
-import "./src/env.mjs";
+import { env } from "./src/env.mjs";
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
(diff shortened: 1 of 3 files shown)


michaelrambeau:
Note for @zawarudo the E2E test suite is failing because it targets the "classic" app, instead of the Next.js one.

image

This is why there is this error (in the classic app, the title of the page is different (`<Project's name> | Best of JS`):

image

PR #228, opened 2023-10-22
fix(e2e:ci): make sure we target the bestofjs-nextjs

zawarudo:
## Goal
Ensure e2e runs on the `nextjs` version of our vercel deployment.

Needed to specify the correct `environment:` in the plugin.

### Gotchas:
Github's environment names use a special ASCII dash

> example: "Preview – bestofjs"

You need to confirm the name as returned from the API

```
curl -L \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer API_AUTH_TOKEN" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/repos/bestofjs/bestofjs/deployments
  ```

## How to test
When bestofjs-classic vercel builds and deploys first, our e2e test will not run.

You can confirm by the timestamps on this PR's run
by checking `bestofjs-classic` was ready first, and the e2e tests waited until the correct environment was up.

1. Classic was deployed first.
image
image

2. E2E test waited for the correct version of the app to be deployed.
image

### DIFF
--- a/.github/workflows/pull-request-check.yml
+++ b/.github/workflows/pull-request-check.yml
@@ -34,6 +34,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           check_interval: 5 # seconds
           max_timeout: 600 # seconds
+          environment: "Preview - bestofjs" # https://github.com/patrickedqvist/wait-for-vercel-preview/issues/16#issuecomment-1140179571
 
       - name: Code Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


michaelrambeau (approved):
Thank you for looking at this setup so quicky!

I assume the name of the deployment you setup (`Preview – bestofjs`) includes the name of the project in Vercel (`bestofjs` for the Next.js app).

image


Linking the problem we had because of the wrong target: https://github.com/bestofjs/bestofjs/pull/227#issuecomment-1773995965

In the future, it would be great to trigger a deploy only for `bestofjs` Next.js app for the related commits (instead of re-building the two apps)

PR #238, opened 2023-11-01
Truncate with CSS project's description in tables

michaelrambeau:
## Goal

Inside projects tables, truncate the project description using CSS (TailwindCSS `truncate` class) so that it's displayed on one single row.

This is necessary because the description from data (the JSON static API) is limited to 70 chars which is enough for a single row, but now enough for two, leading to a bad layout.

The problem I had previously is that `flex-shrink: 0` was missing in right sidebar (the "Featured projects"), breaking the layout.

This PR should address a comment made by the great @zawarudo using Vercel comments feature: https://vercel.live/link/bestofjs-git-feature-better-og-images-bestofjs.vercel.app?page=%2F%3FvercelThreadId%3D_ZSfg&via=in-app-copy-link

## How to test

- On the home page, check that the description is correctly truncated inside "Hot Projects" and "Latest Additions" sections
- Check some search results pages too

## Screenshots

image

image


zawarudo (approved):
Sweet!

PR #239, opened 2023-11-02
Fix broken layout because of truncated text

michaelrambeau:
## Goal

Following the change made here #238 to truncate the description of the projects with CSS, the table is wider than its container and it breaks the layout!

So we need `max-width: 0` actually 😅 

image

## How to test

Ensure everything is aligned properly in the top page!


zawarudo (approved):
Whoops I missed that, visual regression testing time ;)?

Issue #240, opened 2023-11-02
Investigate JS bundle size


michaelrambeau:
The application relies a lot on React Server Components to ship the least amount of JS code to clients.

Still the JD bundles are big (+100 kB), it's a lot of code and it has an impact on performance.

Could we do better? The first step would be to investigate the composition of the different bundles using a tool like https://www.npmjs.com/package/@next/bundle-analyzer

Check the integration here: https://blazity.com/open-source/nextjs-enterprise-boilerplate

We need to know:

- how much JS we ship in the home page
- what is the baseline (the cost of Next.js and React dependencies) and what is the cost of the client features (mainly related to `shadcn/ui` components)

Bonus: integrate [size-limit](https://github.com/ai/size-limit) package in our pipeline (to get notified any time the bundle size changes?)

Evomatic:
@michaelrambeau I looked into the bundle-analyzer and can provide some screenshots of the output.

**Terminal:** Each value from the 'First Load JS' column is the amount of data downloaded from the server. So bundle size per page. From the looks of it `/feature`, `/projects` and `/tags` are the largest. While the largest JS file is `chunks/main-app-31abea25456e3fbf.js` at `233 B`
![image](https://github.com/bestofjs/bestofjs/assets/27636896/064a1e94-43b5-4c58-916a-535d48d7c31e)

**Client:** 
![image](https://github.com/bestofjs/bestofjs/assets/27636896/a712510f-1c28-4396-a5b0-0d4f836530f1)

**Server:** 
![image](https://github.com/bestofjs/bestofjs/assets/27636896/27dc10d0-0e2d-4abb-a09b-7c0365b5fc64)

With the visuals you can zoom in to see each module when opening the html file in the browser. So unfortunately it might be kind of difficult to see details with the screenshots I provided

michaelrambeau:
Thank you @Evomatic  for the investigation, I think it's worth adding the feature to explore the bundles, with some explanation in the README about how to run the analyze process.

Next.js generates a lot of files so I'd say it's important to focus on the bundle related to the top page (`/`) to get a better understanding.

Just for the records, I'm sharing some numbers I can observe from the DevTools about the biggest JS files:

Transferred | Uncompressed | Content?
--- | --- | ---
55.6 kB | 172 kB | ?
31.7 kB | 119 kB | ?
25.1 kB | 78.4 kB | ?
23.2 kB | 70.8 kB | ?
10.6 kB | 35 kB | ?

(and many more)

[...]

PR #245, opened 2023-11-06
Add next.js bundle analyzer

Evomatic:
## Goal
Add Next.js [bundle-analyzer](https://www.npmjs.com/package/@next/bundle-analyzer) plugin which will allow us to analyze the size and composition of the app's bundle. 

For more details please see https://github.com/bestofjs/bestofjs/issues/240
## How to test

- Run `pnpm analyze`
- Inspect the bundles via terminal output or browser

## Screenshots

Terminal:
image

Browser:
image


Evomatic:
**Notes**
- the package https://www.npmjs.com/package/cross-env which allows us to set environment variables in our `package.json` is in maintenance mode, but is still commonly used in examples I saw online. Including the [next-enterprise ](https://github.com/Blazity/next-enterprise/blob/139ed0bef708c62ddc7e96b80ebe8b1cce708317/package.json#L89C1-L89C1) repo. 
- This might be an issue with my local dev but I am having trouble with the analyzer automatically opening in my browser. I added a note in the README about how to do it manually.

Evomatic:
@michaelrambeau I have added a workflow for the bundle analyzer but it still needs some tweaking. Please take a look when you have chance! 
I used the [template](https://github.com/hashicorp/nextjs-bundle-analysis/blob/main/template.yml) that is provided in the repo for the Github action. 

Also have you heard of [act](https://github.com/nektos/act)? It is a way to test github actions locally through docker. I found this quite useful!

.github/workflows/nextjs_bundle_analysis.yml
```diff
@@ -0,0 +1,133 @@
...
+        with:
+          # if you use a custom build directory, replace all instances of `.next` in this file with your build directory
+          # ex: if your app builds to `dist`, replace `.next` with `dist`
+          path: .next/cache
+          # change this if you prefer a more strict cache
+          key: ${{ runner.os }}-build-${{ env.cache-name }}
+
+      - name: Build next.js app
+        # change this if your site requires a custom build command
+        run: pnpm build-project-data && next build
```
michaelrambeau, line 64: I think you should set up the command `pnpm build` here, instead of repeating what is defined inside `package.json`?

.github/workflows/nextjs_bundle_analysis.yml
```diff
@@ -0,0 +1,133 @@
...
+
+      # Here's the first place where next-bundle-analysis' own script is used
+      # This step pulls the raw bundle stats for the current bundle
+      - name: Analyze bundle
+        run: npx -p nextjs-bundle-analysis report
```
michaelrambeau, line 69: I guess `npx` works too but it may be more consistent to put the PNPM command instead? (`pnpm exec` maybe)

Evomatic:
So I made some adjustments, using the `pnpm` command for consistency and added configs for the bundle analysis in `package.json` as mentioned in the [README](https://github.com/hashicorp/nextjs-bundle-analysis/blob/main/README.md) for the bundle analysis. I got the github workflow successfully running locally but it still fails after pushing to github (as you can see). Looks like the issue is with artifact generation and retrieval from Upload bundle step. the file at path `.next/analyze/__bundle_analysis.json` is not being found. Not sure of the exact issue yet. Will continue with the investigation!

Evomatic:
@michaelrambeau I decided to remove the bundle analysis workflow for now as the issue with downloading artifacts that are uploaded is still unresolved. As you mentioned before, checking manually will suffice for now and I may come back to this at a later point. I can add it as a issue for future reference.

michaelrambeau (approved):
Thank you very much @Evomatic I've checked again the command `pnpm -F bestofjs-nextjs analyze` and it works well!

Like we said on Discord, we don't really need an automated workflow that checks the bundle size after every single commit, as long as we do ii from time to time, when upgrading or adding dependencies, to ensure the bundles sent to the client stay "in control".

A cool feature I've just found: I wanted to check the impact of 2 dependencies I've added recently (see #254 ):

- `unstated-next`
- `use-debounce`

In the left sidebar, you can enter the name of the package in the box `search modules` and it will be highlighted in red!
For `unstated-next` it's so small that you have to zoom to see the content of the red box (this is the reason I was not able to find it when I was checking the different bundles)

image

PR #247, opened 2023-11-11
Improve search UX merging projects and tags

### LINKED ISSUE #246: Improve search UX
Following a discussion with @zawarudo , the most relevant results don't always show up first.

If you search for "webrtc", the first entry is a project call `PeerJS` because it's the most popular **project** that has the string `webrtc` in its name or description. If you press key down and Enter, the user would be redirected to the `PeerJS` project page.

It would make more sense to open the page related to the **tag** `Web RTC` instead: https://bestofjs.org/projects?tags=webrtc

Image

## How it works now

The problem is the search feature prioritizes the search by project over the search by tags: search results are segregated between "projects" and "tags" and we show projects first because I wanted users to find easily the projects they are looking for.

E.g.: type `react`, push the down key and Enter and you go to React project directly.

I use the search feature as a fast way to access either the repository or the home page, for projects I know I can find on _Best of JS_.

The use case mentioned above is more about exploring the eco-system rather than looking up a specific project.

> When I query and hit enter I get search results.
> Then I can take my time and have more real estate to make a decision which project I want to research first

How to satisfy both needs?

## Merge projects and tags?

My first idea would be to merge projects and tags into the same list, ranking them to show the most relevant results first, whether it's a project or a tag.

However it could be disturbing to users:

> I'm seeing a lot of gotchas around applying that to the "default" action when pressing enter

michaelrambeau:
## Goal

Improve the UX when searching, to fix #246 by combining projects and tags in the same list, sorted by relevance, instead of separating them.

To help differentiate them, 2 colors were assigned, from the `Radix Colors` scales:

- orange for projects
- pink "crimson" for tags

We could generalize the use of these colors in other places?

About colors and web apps, I love the `Radix Colors` project: https://www.radix-ui.com/colors/docs/palette-composition/scales

## How to test

Search for "WebRTC" and tell me if you are confused or not 😄 

## Screenshots

image

image

### DIFF
--- a/apps/bestofjs-nextjs/src/app/globals.css
+++ b/apps/bestofjs-nextjs/src/app/globals.css
@@ -6,6 +6,34 @@
   /* Light values come from the "stone" theme, except for the primary colors */
   /* https://ui.shadcn.com/themes */
   :root {
+    /* `orange` scale */
+    --orange-1: #fefcfb;
+    --orange-2: #fff7ed;
+    --orange-3: #ffefd6;
+    --orange-4: #ffdfb5;
+    --orange-5: #ffd19a;
+    --orange-6: #ffc182;
+    --orange-7: #f5ae73;
+    --orange-8: #ec9455;
+    --orange-9: #f76b15;
+    --orange-10: #ef5f00;
+    --orange-11: #cc4e00;
+    --orange-12: #582d1d;
+
+    /* `crimson` scale from Radix colors */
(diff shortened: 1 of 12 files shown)


michaelrambeau:
## Making everything bigger

A bit inspired by TailwindCSS except we don't invert colors for the highlighted item in the light mode

image

image

michaelrambeau:
Closing as I made a another PR to improve the search UX #253

PR #250, opened 2023-11-16
Introduce Radix Colors scale and improve the design

michaelrambeau:
## Goal

Improve the design by introducing the color scales from https://www.radix-ui.com/colors by adding the CSS variables for the following colors:
  - `sand`
  - `orange` 

The color tokens were simply copy-pasted from the _Scales_ page:

image

Main benefit:
- the same CSS variables can be used for both light and dark mode
- more consistency
- the scale is really well documented

Other changes:

- Change GitHub and Discord icons to use an "outline" style
- Remove hard-coded TailwindCSS colors

## How to test

Check the look-and-feel in both dark and light mode:

- top header links
- project links

## Screenshots

image

image


michaelrambeau:
## More changes

- Add a subtile gradient in table headers, in dark mode only
- make prev/next buttons in "Featured" section more visible
- improve tags (current tags and related tags) in `/projects?` search pages

image

image

flamrdevs:
I had a catch. some Tailwind classes don't work properly. This is because the color variable is in hex format, so the [opacity modifier syntax](https://tailwindcss.com/docs/text-color#changing-the-opacity) does not work.

solutions :
- remove all opacity modifier syntax classes that refer to hex colors
- change color variable from hex to rgb or hsl (Recommended)

[Tailwind - Using CSS Variables](https://tailwindcss.com/docs/customizing-colors#using-css-variables)

michaelrambeau:
Thank you @flamrdevs for the comment.

Yes, I was aware that the reason colors were defined as HSL in shadcn/ui values was to be able to adjust the opacity using the TailwindCSS modifiers (E.g. `bg-background/50`). However, if we use a whole color scale well defined such as the one provided by Radix Colors (instead of the few colors defined by shadcn/ui), I thought it was better to only pick the tokens instead of adjusting the color opacity by ourselves in the markup, this the reason I didn't convert the Radix Colors from RGB to HSL.

In the codebase the only HSL remaining colors are `secondary` and `destructive` but I don't use these variations, we could remove them.

michaelrambeau:
Also, after having checked [Radix Themes](https://www.radix-ui.com/), I'm very tempted to introduce it as it seems they do everything right, it could make the code cleaner (especially thanks to `Box` and`Flex` utilities).

Of course everything is a tradeoff as everything is easy to customize with shadcn/ui components but I like very much how robust Radix Theme component look, I wonder if you have tried them?

flamrdevs:
> Yes, I was aware that the reason colors were defined as HSL in shadcn/ui values was to be able to adjust the opacity using the TailwindCSS modifiers (E.g. `bg-background/50`). However, if we use a whole color scale well defined such as the one provided by Radix Colors (instead of the few colors defined by shadcn/ui), I thought it was better to only pick the tokens instead of adjusting the color opacity by ourselves in the markup, this is the reason I didn't convert the Radix Colors from RGB to HSL.

So the conclusion is that Radix Colors will still use the hex format. right? And the next step is to delete the class that uses the Opacity Modifier which refers to Radix Colors.

> In the codebase the only HSL remaining colors are `secondary` and `destructive` but I don't use these variations, we could remove them.

Not only is it secondary or destructive variant, this [example class](https://github.com/bestofjs/bestofjs/blob/8293ae3e03e50f869daffd5730dfbcb0137637bc/apps/bestofjs-nextjs/src/components/ui/button.tsx#L17C54-L17C73) also won't work, because the color in [tailwind.config.js](https://github.com/bestofjs/bestofjs/blob/8293ae3e03e50f869daffd5730dfbcb0137637bc/apps/bestofjs-nextjs/tailwind.config.js#L23C11-L23C37) refers to [radix colors](https://github.com/bestofjs/bestofjs/blob/8293ae3e03e50f869daffd5730dfbcb0137637bc/apps/bestofjs-nextjs/src/app/globals.css#L42C5-L42C33) which uses hex format.

flamrdevs:
> Not only is it secondary or destructive variant, this [example class](https://github.com/bestofjs/bestofjs/blob/8293ae3e03e50f869daffd5730dfbcb0137637bc/apps/bestofjs-nextjs/src/components/ui/button.tsx#L17C54-L17C73) also won't work, because the color in [tailwind.config.js](https://github.com/bestofjs/bestofjs/blob/8293ae3e03e50f869daffd5730dfbcb0137637bc/apps/bestofjs-nextjs/tailwind.config.js#L23C11-L23C37) refers to [radix colors](https://github.com/bestofjs/bestofjs/blob/8293ae3e03e50f869daffd5730dfbcb0137637bc/apps/bestofjs-nextjs/src/app/globals.css#L42C5-L42C33) which uses hex format.

You can check this in [hall-of-fame](https://bestofjs.org/hall-of-fame) page
image
The hover style of Search button not as expected

flamrdevs:
> Of course everything is a tradeoff as everything is easy to customize with shadcn/ui components but I like very much how robust Radix Theme component look, I wonder if you have tried them?

Since I built [Klass](https://klass.pages.dev), I completed the design of this system. In my personal web, I can have Box, Flex, Grid, Container, and other components that use a style system like the one used in [Radix Themes](https://www.radix-ui.com/themes/docs/theme/layout), [Chakra UI](https://v1.chakra-ui.com/docs/styled-system/features/responsive-styles), and [Mantine UI](https://mantine.dev/styles/style-props).

michaelrambeau:
Thank you very much @flamrdevs for having checked and sorry for the misses!

The primary button used in the Hall of Fame uses the syntax `hover:bg-primary/90` to decrease the opacity, when hovered, my bad, I'm fixing...

PR #252, opened 2023-11-18
Cleanup CSS variables, add missing `:hover` effect

michaelrambeau:
## Goal

Cleanup following the introduction of Radix Colors tokens #250 , taking into account the comment from @flamrdevs as there was a miss about the `:hover` color effect of some buttons.

- The 2 variants `primary` use  `secondary` use tokens from the `orange` and the `sand` scales respectively
- The `destructive` variant was removed
 
Also new CSS variables are introduced for links:

- `link-foreground`
- `link-underline`

## How to test

- Go to the Hall of Fame `/hall-of-fame`
- Check the `:hover`effect of the Search button, in both light and dark mode

## Screenshots

image

### DIFF
--- a/apps/bestofjs-nextjs/src/app/globals.css
+++ b/apps/bestofjs-nextjs/src/app/globals.css
@@ -49,8 +49,6 @@
     --muted-foreground: var(--sand-11);
     --accent: var(--sand-2);
     --accent-foreground: black;
-    --destructive: 0 84.2% 60.2%;
-    --destructive-foreground: 60 9.1% 97.8%;
     --border: var(--sand-4);
     --input: var(--border);
     --ring: var(--orange-3);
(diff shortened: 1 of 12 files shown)


flamrdevs:
Looks better now, nice work.

flamrdevs:
I think using [color scale 3](https://github.com/bestofjs/bestofjs/blob/8293ae3e03e50f869daffd5730dfbcb0137637bc/apps/bestofjs-nextjs/src/app/globals.css#L54) for the `ring` color is too dim, I have difficulty finding which component is in focus state, maybe it's better to use color scale 9.

michaelrambeau:
> I think using [color scale 3](https://github.com/bestofjs/bestofjs/blob/8293ae3e03e50f869daffd5730dfbcb0137637bc/apps/bestofjs-nextjs/src/app/globals.css#L54) for the `ring` color is too dim, I have difficulty finding which component is in focus state, maybe it's better to use color scale 9.

Thank you @flamrdevs ! Fixed here https://github.com/bestofjs/bestofjs/pull/252/commits/4d51e725fe6cb7406543273301d4d72a990632a4

PR #253, opened 2023-11-22
Combine projects and tags in the same search results list

### LINKED ISSUE #246: Improve search UX

michaelrambeau:
## Goal

- Improve the UX when searching, to fix #246 by combining projects and tags in the same search result list, sorted by relevance, instead of separating them.
- When a tag is selected, make the fact that the search is scoped to the tag mroe obvious

## How to test

- Try to search for `react`, notice how the list shows both the React project and the `React` tag
- Open a tag (say the `React` tag), then open the search modal

#### Default state (no query, no tags)

Showing the most popular tags

image

#### Searching

Showing a mix of projects and tags

image

#### Start the search when a tag is selected

- The place holder has changed to explain that the search is now "scoped" to one or several tags
- The list shows the "Related Tags" instead of the popular tags
- The icon of the tag entry has changed to convey the fact that it will open a page that combine several tags (instead of a single one)

image

#### Searching when a tag is selected

image

### DIFF
--- a/apps/bestofjs-nextjs/package.json
+++ b/apps/bestofjs-nextjs/package.json
@@ -25,6 +25,7 @@
   "dependencies": {
     "@klass/core": "^3.4.8",
     "@klass/react": "^3.4.8",
+    "@radix-ui/react-checkbox": "^1.0.4",
     "@radix-ui/react-collapsible": "^1.0.3",
     "@radix-ui/react-dialog": "^1.0.4",
     "@radix-ui/react-dropdown-menu": "^2.0.5",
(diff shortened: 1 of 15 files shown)


michaelrambeau:
## Next iteration

Adding checkboxes to toggle projets and tags

image

PR #254, opened 2023-11-24
Improve search palette feature (debounce + refactoring)

michaelrambeau:
## Goal

- Debounce calls to search functions in the Search palette to avoid blocking the UI while the user is typing
- Refactor the Search Palette feature, to split state and components, and introduce a library [Unstated Next](https://github.com/jamiebuilds/unstated-next) that is a thin wrapper around React Context API (unfortunately nobody knows it, but I like it very much)
- Print timing information in debug mode to measure the time it takes to perform searches
- Make search algorithm a bit faster by avoiding creating the same RegExp objects for each project the function loop through when ranking every project (about 2000 projects)
- Add basic tests for search functions

## How to test

Type very fast in the search input and ensure the search feature works!

Test command: `pnpm exec vitest src/lib/search-utils.test.ts`

image

## Screenshot

image


apps/bestofjs-nextjs/src/lib/search-utils.tsx
```diff
@@ -1,86 +1,97 @@
+const DEBUG_MODE = false; // to show time spent in search functions
```
zawarudo, line 1: // TODO: add to .env

😄 ?

apps/bestofjs-nextjs/src/components/search-palette/search-palette-state.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  const onOpenChange = (value: boolean) => {
+    if (!value) {
+      resetCurrentTags();
+      setSearchQuery("");
+    }
+    setOpen(value);
+  };
+
+  React.useEffect(() => {
```
zawarudo, line 85: We might want to add a comment summarizing this useEffect.

Or refactor the logic outside into a function like `_enableSearchPaletteShortcutListener`

zawarudo (approved):
Works well,

Left some discussion points in the comments

apps/bestofjs-nextjs/src/components/search-palette/search-palette-state.ts
michaelrambeau: Good point, each non-trivial `useEffect` should be extracted in a custom hook, `useKeyboardShortcut()` for example.

apps/bestofjs-nextjs/src/lib/search-utils.tsx
michaelrambeau: Thank you, I've added an environment variable accessible from the client: https://github.com/bestofjs/bestofjs/pull/254/commits/674b618630860c79545694b3f241c964ba74b35b

Issue #217, opened 2023-10-16
Add "hover cards" to get a quick overview of projects and tags


michaelrambeau:
## The need

I noticed a nice pattern in GitHub: when a repository, a commit or a user is mentioned, by hovering the link it's possible to get an overview about the link in a popup window.

Examples:

#### Repository link from the default feed

![image](https://github.com/bestofjs/bestofjs/assets/5546996/a8a8d6e3-82e9-491d-be01-820bcb4c60ae)

#### GitHub avatar and username from an issue

![image](https://github.com/bestofjs/bestofjs/assets/5546996/e8fccdbe-e6e7-46c0-b426-0dca71ec4485)

#### Link to PR from PR list

![image](https://github.com/bestofjs/bestofjs/assets/5546996/e5bb7303-e4ef-4b22-84d9-4066a91f0a3a)

#### Commit link from a comment

![image](https://github.com/bestofjs/bestofjs/assets/5546996/985bb46b-7406-41fd-a3c8-95299b3b23e4)

#### In Discord

I have seen a similar pattern in some Discord servers (the Next.js one for example): you get an overview of the content of a channel when you hover a channel name, from the left sidebar:

![image](https://github.com/bestofjs/bestofjs/assets/5546996/6ab9c9b4-e1bd-45f1-ad0a-ed1c9c943ec3)

I'd like to see the same feature in _Best of JS_, especially for projects and for tags, as the definition and the scope of tags is not often obvious.

It should be easy to implement as `shadcn/ui` components include "hover cards": https://ui.shadcn.com/docs/components/hover-card

## What to show

When hovering a tag link, I'd like to see:

- the tag description
- a list of the 5 most popular projects included
- a link to the tag page

michaelrambeau:
At first, I think the cards would be useful for the tag links we have inside each project row, in every project list, because the meaning of the tags is not not always obvious.

The card would show:

- a summary of the other projects available under the tag
- a description of the tag (I need to make the tags description available in the API)

Evomatic:
That is understandable. For now I will focus on making it viable for each project row on the `/projects` page. 

Here is a demo of it in action!

https://github.com/bestofjs/bestofjs/assets/27636896/8540efa6-4dd8-4da0-9042-84094226c848

Evomatic:
[...]

PR #257, opened 2023-11-30
Add hover cards on project tag links

Evomatic:
## Goal
Add hover cards to tag links in order to get an overview of the link and to show the top 5 projects. 
More details about this feature can be found in the issue: https://github.com/bestofjs/bestofjs/issues/217

## How to test

- Simply hover over a tag to see it appear.


Evomatic:
@michaelrambeau I now have a working hover card that will show the following:

- Tag icon followed by the tag name
- Tag description
- "TOP 5 projects" followed by a list of projects icon and name

The styling still needs a bit of work but this at least gives us an idea of the overall "feel" of having this feature!

michaelrambeau:
> @michaelrambeau I now have a working hover card that will show the following:
> 
>     * Tag icon followed by the tag name
> 
>     * Tag description
> 
>     * "TOP 5 projects" followed by a list of projects icon and name
> 
> 
> The styling still needs a bit of work but this at least gives us an idea of the overall "feel" of having this feature!

Thank you Evan, that's a great progress!

It seems the build failed because of a TS error: https://vercel.com/bestofjs/bestofjs/Hcotzs4coRGwFYrCyeBmePoG1Mx7

```
Type error: Property 'projects' is missing in type 'Tag' but required in type 'TagWithProjects'.
```

We can have a look together as it's sometimes tricky to get the types right.

Evomatic:
> Thank you Evan, that's a great progress!
> 
> It seems the build failed because of a TS error: https://vercel.com/bestofjs/bestofjs/Hcotzs4coRGwFYrCyeBmePoG1Mx7
> 
> ```
> Type error: Property 'projects' is missing in type 'Tag' but required in type 'TagWithProjects'.
> ```

Thank you for the feedback! I fixed the type error with my latest commit. The issue was I was setting the param type as `TagWithProjects` in components that don't expect tag with project data. Happy to chat still if you have more ideas! 

For the design of the hover card, I loosely followed the design used on the OG images, as the familiarity/consistency seemed appropriate. Obviously there is more worked that needs to be done, so this is something we should discuss!

michaelrambeau:
Thank you Evan, the hover cards on the preview URL look great: https://bestofjs-git-add-hover-cards-bestofjs.vercel.app/

image

The client-side caching works well, as going to back to a tag that was already visited is "instant".

In the PR diff, do you know what happened to `project-avatar.tsx`, it looks like the whole file was changed but I cannot see the difference?

I'm going to see if we can improve the look and feel of the cards.

apps/bestofjs-nextjs/src/components/tags/project-tag-hover-card.tsx
```diff
@@ -0,0 +1,147 @@
...
+    color: "white",
+    gap: 10,
+    alignItems: "center",
+    borderStyle: "solid",
+    borderColor: `rgba(255, 255, 255, 0.2)`,
+    borderBottomWidth: "1px",
+    borderTopWidth: isFirst ? 1 : 0,
+    padding: "8px 0",
+  };
+}
```
michaelrambeau, line 98: I guess it's just for the prototype as I can see you re-used what we did for OG images generated on the server but we should not use inline styles in the web app, especially because we support light and dark modes.

apps/bestofjs-nextjs/src/server/api-tags.tsx
```diff
@@ -97,6 +97,34 @@ export function createTagsAPI({ getData }: APIContext) {
...
+          projectCollection,
+          searchQuery
+        );
+
+        tag.projects = projects.map(populate);
+      }
+      return {
+        tags,
+      };
+    },
```
michaelrambeau, line 127: Since the slug is a unique key, I think it will be better if this API method returns a single tag instead of an array of one one tag: `return { tag: }`.

Evomatic:
> In the PR diff, do you know what happened to `project-avatar.tsx`, it looks like the whole file was changed but I cannot see the difference?

In `project.avatar.tsx` I was placing console logs when I was exploring how the component works. At some point I might of ran "ran format document" which may be the reason for the large difference. Sorry, I don't know the exact reason for this change. I could attempt to "cherry pick" this commit and remove this specific change 🤔

apps/bestofjs-nextjs/src/server/api-tags.tsx
Evomatic: Understood, I will look into fixing this.

apps/bestofjs-nextjs/src/components/tags/project-tag-hover-card.tsx
Evomatic: In this case is it better to stick with tailwind utility class names then?

Evomatic: Since you mentioned wanting to try and improve the look and feel on your end, I will leave this part up to you for now 😄

michaelrambeau:
## Refactor and improve the layout

I did some refactoring in the latest commit https://github.com/bestofjs/bestofjs/pull/257/commits/82c793f8f6bea8120b9df16c0fa289baf403174b

The main change: we don't need to manage an `isOpen` state to trigger the fetch only when a hover card is displayed on the screen

Also I changed the layout:

- add links on the projects
- add a link `View all` at the bottom
- ensure there is no layout shift when transistioning from the skeleton to actual data
 
image

michaelrambeau (approved):
Let's ship this nice feature to production!

michaelrambeau:
Doing more tests and after having talked with @zawarudo we noticed it can be hard to discover the feature because the hover card shows up only after 700ms.

We can make the hover cards "instant" but then it becomes obstrusive, the hover card shows up "by accident" when you just want to reach a link at the bottom of the screen for example.

Asking AI:

>  While there is no one-size-fits-all answer, it is generally recommended to introduce **a slight delay** before showing the hover card to avoid it being too sensitive to minor mouse movements. This can help prevent the hover card from being displayed when the user is simply passing over a link. Additionally, the hover card should appear **quickly enough** to provide a seamless and responsive experience for the user.

So there is a balance to find: not too slow but not too fast 😅 !

Settings are described here:

https://www.radix-ui.com/primitives/docs/components/hover-card#api-reference

- `openDelay`: 700ms by default
- `closeDelay`: 300 ms by default

Evomatic:
> Doing more tests and after having talked with @zawarudo we noticed it can be hard to discover the feature because the hover card shows up only after 700ms.
> 
> We can make the hover cards "instant" but then it becomes obstrusive, the hover card shows up "by accident" when you just want to reach a link at the bottom of the screen for example.
> 
> Asking AI:
> 
> > While there is no one-size-fits-all answer, it is generally recommended to introduce **a slight delay** before showing the hover card to avoid it being too sensitive to minor mouse movements. This can help prevent the hover card from being displayed when the user is simply passing over a link. Additionally, the hover card should appear **quickly enough** to provide a seamless and responsive experience for the user.
> 
> So there is a balance to find: not too slow but not too fast 😅 !
> 
> Settings are described here:
> 
> https://www.radix-ui.com/primitives/docs/components/hover-card#api-reference
> 
>     * `openDelay`: 700ms by default
> 
>     * `closeDelay`: 300 ms by default

Good catch! I think decreasing the delay will make everything feel less "sluggish" or "heavy", providing a better experience!

michaelrambeau:
## Impact of bundles

It seems the Hover Card component from Radix adds 13 kB to the bundle (TODO: check bundle analyzer)

image

Issue #267, opened 2024-01-07
Typewriter effect on top page


michaelrambeau:
To make _Best of JS_ page fun and show the variety of domains we cover, I'd like to see a typewrite effect applied on main heading ("The best of JS, HTML, CSS" currently)

The cycle could be:

- Show: `the best of JavaScript`
- Several backspace keystroke to remove the word JavaScript
- keystrokes to add the word `TypeScript`
- remove the word TypeScript
- Add `React`
- And so on with some popular topics like `Vue.js`, `Astro`, `Bun`, `Deno`, `CSS-in-JS`...

Example of projects we could use: https://bestofjs.org/projects/typeit

michaelrambeau:
About the implementation, we should check what Drizzle uses in their top page, because it's exactly the kind of effect I want!

https://orm.drizzle.team/

![Image](https://github.com/bestofjs/bestofjs/assets/5546996/5b1f2632-59dc-4e83-889a-09d53303eb00)

Evomatic:
@michaelrambeau I am interested in taking a look into this unless you wanted to 😄

PR #268, opened 2024-01-17
Typewriter effect

Evomatic:
## Goal

This is to add a component that creates the typewriter effect that you can see on websites such as https://orm.drizzle.team/
The component was built without using any packages, giving us more flexibility for how we want to use it and for any future refactoring. I loosely followed Drizzle's implementation and watched/read a few tutorials before trying to create my own.

One idea I had for the typewriter that I thought could be interesting, is that currently there is a hard coded array of what topics are to be passed to the typewriter, but I wonder if it worth while to fetch and populate the array with the top topics dynamically? For example if we wanted to cycle through hot topics of the day. This potentially could slow down load time of the page though... 🤔 

resources:

- https://github.com/drizzle-team/drizzle-orm-docs/blob/5c113a5a732ca106dda5ca7dd51995e35eea87e1/src/components/LandingPage/Header/Typer.astro

- https://www.youtube.com/watch?v=R846J8LJ6os&t=644s  

Related issue: https://github.com/bestofjs/bestofjs/issues/267
## How to test
Checkout it out here: https://bestofjs-git-typewriter-effect-bestofjs.vercel.app/

## Screenshots

I ran a bundle analysis but didn't notice an increase. 
image

desktop:
typewriterSEO

mobile:
image


michaelrambeau:
That is beautiful. thank you so much Evan! 😻 

I've checked dark and light modes, in Firefox and Chrome, it's perfect!

I love the fact no extra package is introduced.

About the topics, I cannot really see the need of fetching hot topics for now, especially because not all topics are related to a specific tag but it something we could do in the future.

Maybe we could add "Node.js" in the list, at a high position, just after "TypeScript"?

apps/bestofjs-nextjs/src/components/home/typewriter.tsx
```diff
@@ -0,0 +1,80 @@
...
+        className="underline decoration-[var(--logo-color)]"
+        ref={typeWriterRef}
+        id="typewriter"
+      ></span>
+      <span className="animate-cursor-pulse" id="cursor">
+        |
+      </span>
+    </div>
+  );
+}
```
michaelrambeau, line 80: I wonder if we really need `useState` and `useEffect`?

Also I think the `id` in the markup ( (`id="typewriter"` and `id="cursor"`) are not used?

Evomatic: @michaelrambeau I was able to remove `useState` and `useEffect` so now we are just relying on `useRef`

Evomatic:
@michaelrambeau Ran into a bug I am not sure about so maybe you have an idea!

When navigating between pages and going back to the home page, the typewriter begins to "jitter and dance" for lack of a better word XD. I tried to diagnose is but not really sure what could be causing this bug.

michaelrambeau (commented):
Thank you it works well now when navigating and returning to the homepage.

A small typo: `milliseconds` takes 2 `l` (I use the extension `Code Spell Checker` in VSCode to spot typos!)

When you set up a timer, it's usually a good practice to remove it in the cleanup function of the `useEffect`.

It's a small detail but I would re-order the functions, so that the reading of the code is more natural.

- starting  with the `useEffect` call, as it's the starting point
- and then the most important function `animateText`
- finally the least important functions

michaelrambeau (approved):
Thank you so much, let's ship it 🚀

PR #274, opened 2024-03-19
Add time-period sort picker on Homepage onto Hot Projects list 

### LINKED ISSUE #270: This week / this month rankings from the home page
Written by Michael:

The previous version of _Best of JS_, still available at https://classic.bestofjs.org/, provides a way to check quicky the trends from the last 7 days, the last 30 days and the last 12 months, right from the home page

Image

We could implement it using URL segments to avoid managing the state:

- `/` by default the top page shows the "Hot Projects Today"
- `/this-week` would show the "Hot projects this week" (by number of stars added the last 24 hours)
- `/this-month`
- `/this-year`

SaulSilver:
Resolves #270 

## Goal
Provide a quick way to check the trends from the last 7 days, the last 30 days and the last 12 months, right from the home page.

Please see the commit messages to follow the changes.

### Things to consider
1. I noticed the selector doesn't only change the hot projects list, but it also affects the "Featured" projects list. This behavior also occurs on the older website version https://classic.bestofjs.org/, I thought this behavior is misleading the user UI-wise, because the selector is within the "Hot projects" list container.
2. In the issue description, the naming convention for the URL is e.g. `/this-week` that would show the "Hot projects this week". However, I opted to use "weekly" in the PR to follow the same naming convention used for the sorting of other projects lists (e.g. https://bestofjs.org/projects?sort=weekly)

## How to test
- Run `pnpm dev` OR open the preview link posted in Vercel's bot msg below
- Try any of the options on the Hot Projects selector on homepage
- Notice the change of the projects in the list and the URL - the query string should follow the option selected. FYI: the default option shows the daily so `/` and `/?sort=daily` are the same

## Screenshots

Screenshot 2024-03-19 at 13 01 52

---

Screenshot 2024-03-19 at 13 02 12


michaelrambeau:
Thank you so much Hatem for the feature!

I works well but I noticed there is a short delay when switching between the time ranges.
Could it be improved by used path parameters instead of query string parameters?
(in terms of file structure, that means a new page `/[time-range]/page.tsx` that would use shared component with the root `/page.tsx`)

The reason: when using search parameters, Next.js pages are "dynamic".

But with path parameters we can generate statically some page at build times, telling Next.js the what parameters to apply.

The home page (`/`) is rebuilt every day as I trigger a build after data is updated every day.
We would rebuild the new pages too: `/this-week`, `this-month` and `/this-year`.

Another thing: the link "View full rankings" at the bottom of the "Hot projects" card should be updated to reflect the current time range.

I'm aware that the way you implement is very consistent with how the projects page work (with the "sort" parameter), that is a great point, so I'm sorry to ask you to explore this approach!

About E2E failing, I'm sorry too because a few days there was a very unexpected build error on Vercel, I had a hard time debugging, at some point I had to disable the Playwright config file that was causing a compilation error but I didn't have the time to go back to the issue, my bad.

michaelrambeau:
Thank you again Hatem @SaulSilver and sorry for the time it took me to handle this feature here: https://github.com/bestofjs/bestofjs/pull/352

PR #275, opened 2024-03-21
Move LICENSE file to root directory & add Code of Conduct file

SaulSilver:
I also updated the LICENSE copyright year.


michaelrambeau (approved):
Thank you Hatem, that makes sense, I forgot to move some files when the monorepo setup was introduced last year.

PR #277, opened 2024-03-23
Restore end-to-end test suite with Playwright

michaelrambeau:
## Goal

Restore the E2E test suite that I had to disable 2 weeks ago, when the build was broken on Vercel.

I'm not too sure what happened at the time, there was no change from our side but one day the project was not building anymore: 

the error was a TypeScript error because of `playwright.config.ts` file, the package `@playwright/test` was not found (which is normal since we run in production mode on Vercel, devDependencies are not installed as we want the test suite to run on GitHub actions only)

So this PR restores the test suite by:

- restoring the config file
- ensure the TS compiler only compiles files from the `src` folder (maybe I should exclude `playwright.config.ts` instead?)

## How to test

Ensure all the checks pass on GitHub actions!

## Screenshots

image

### DIFF
--- a/apps/bestofjs-nextjs/playwright.config.ts
+++ b/apps/bestofjs-nextjs/playwright.config.ts
@@ -1,0 +1,72 @@
+import { defineConfig, devices } from '@playwright/test';
+
+const PORT = process.env.PORT || 3000;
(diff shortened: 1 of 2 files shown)


zawarudo (approved):
LGTM, unsure what happened but call me when it happens again so I can understand where the problem lies.

PR #280, opened 2024-04-28
Fix open a tag from the search palette when the current page is > 1

michaelrambeau:
## Goal

Fix the bug about the search palette mentioned here: https://github.com/michaelrambeau/bestofjs/issues/938

Cause: the page parameter should be reset when opening a tag from the search palette.

Note: we try to keep the search parameters to maintain the search order picked by the user (E.g. "by number of stars")

## How to test

- Go to "All projects" page
- Scroll and go to the next page to open http://localhost:3000/projects?sort=total&page=2
- Press Cmd K to Open the search palette
- Search for a tag, select it, and check the tag page open correctly

## Screenshots

image


zawarudo (approved):
LGTM

Did a few tests on the deployed URL and it works as expected.

PR #285, opened 2024-07-11
Fix theme toggle

stefanofa:
## Goal
The `theme` field returned by the `useTheme` hook of [next-themes](https://github.com/pacocoursey/next-themes?tab=readme-ov-file#usetheme-1) has the literal value `system` if it has not been manually modified.
So if the user has `light` as the system preference, on the first click the handler 
```ts
onClick={() => setTheme(theme === "light" ? "dark" : "light")}
```
will set the theme to `light` again, since `"system" !== "light"`.

## How to test
Go to https://bestofjs.org/ from an incognito tab with system preference set as light and click the theme toggle icon. On the first click the theme will not change.

## Screenshots
image

### DIFF
--- a/apps/bestofjs-nextjs/src/components/theme-toggle.tsx
+++ b/apps/bestofjs-nextjs/src/components/theme-toggle.tsx
@@ -8,13 +8,13 @@ import { Button } from "@/components/ui/button";
 import { MoonIcon, SunIcon } from "./core";
 
 export function ThemeToggle() {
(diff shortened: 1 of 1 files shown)


michaelrambeau (approved):
Thank you very much @stefanofa great catch!

Speaking of theme toggle, having only the choices between "dark" and "light" is often considered as a bad practice as some user prefer to pick the "system", which is useful when the system itself changes between light and dark mode depending on the time of the day, it's something we could improve.

PR #286, opened 2024-07-13
Allow to select system as preference for themes

stefanofa:
## Goal
Enable to select system as preference for themes.
As suggested in https://github.com/bestofjs/bestofjs/pull/285#pullrequestreview-2176239342

I also took poetic license to add a grow animation on the theme change, which uses view transitions if the browser being used supports them.
It also follows user preferences by avoiding the animation in case the user has `prefers-reduced-motion` set.

If the animation is not liked, the first commit can be considered.

## How to test
As in the video.

## Screenshots
https://github.com/user-attachments/assets/9e624f3d-9587-4e80-afee-71211195bc90


michaelrambeau:
Hello @stefanofa thank you very much for going deep with the theme switcher topic 😄 

And sorry again for the late reply, I'm working on a big migration of the backend and I missed this PR, shame on me!

The animation looks really cool in the video but I'm a bit worried by the complexity it adds to the codebase, considering users will probably not change the theme often (especially with this addition of the "system" option).

The previous animation was a simple one done with CSS and "stolen" from shadcn/ui but they went the same way as us, adding a "system" option and there is no more animation: https://ui.shadcn.com/

So I'd grateful if you could discard the second commit (including the changes in the lock file), thank you for your understanding! 🙏

stefanofa:
Hey @michaelrambeau, I've reverted the second commit as requested, including the changes in the lock file. If there's anything else you need, feel free to let me know.

michaelrambeau:
Thank you so much @stefanofa the feature is live!

And sorry again, I felt bad after having rejected the animation you crafted, using your right to _the poetic license_ (I love that expression!)

Feel free to suggestion any other improvement, now that you are familiar with the codebase.
(and don't hesitate to join us on Discord if you have questions, it's a very quiet server 😅 )

PR #305, opened 2024-08-27
Exclude specific projects from rankings

michaelrambeau:
## Goal

Following discussions we had with @BBlackwo we'd like to promote some projects that don't match the usual criteria and we have created a new tag called "Editor's pick" whose defintion could be:

> Projects that don't match the usual criteria about being either written in a language that compile to JavaScript or a build tool that emits JavaScript, picked by Best of JS authors.

The discussion started with the request to add the 2 following projects:

- Coolify: https://github.com/michaelrambeau/bestofjs/issues/964
- Immich: https://github.com/michaelrambeau/bestofjs/issues/980

However we don't want these projects to show up in the home page "Today's hot projects" because they tend to be the hottest projects every day.

So we want to exclude them from the rankings (daily, monthly, and yearly in the _Rising Stars_), the same way we don't include the learning resources in the rankings.

## Screenshots

image

The tag's description, showing up when "hovering" the tag:

image

### DIFF
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -26,6 +26,7 @@
     "debug": "^4.3.4",
     "dotenv-cli": "^7.4.2",
     "drizzle-orm": "^0.31.2",
+    "es-toolkit": "1.16.0-dev.477",
     "fs-extra": "^11.1.1",
     "p-map": "^7.0.2",
     "p-throttle": "^6.1.0",
--- a/apps/backend/src/tasks/build-static-api.task.ts
+++ b/apps/backend/src/tasks/build-static-api.task.ts
@@ -1,3 +1,5 @@
+import { orderBy } from "es-toolkit";
+
 import { schema } from "@repo/db";
 import {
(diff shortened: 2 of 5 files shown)


apps/backend/package.json
```diff
@@ -26,6 +26,7 @@
     "debug": "^4.3.4",
     "dotenv-cli": "^7.4.2",
     "drizzle-orm": "^0.31.2",
+    "es-toolkit": "1.16.0-dev.477",
```
BBlackwo, line 29: Should we be using the `-dev.477` version? We probably just want `1.16.0`, right?

BBlackwo (approved):
Looks good, just one small comment. Nice work 🚀

apps/backend/package.json
michaelrambeau: I had a TS issue with the the `1.16.0`, the types not being up to date.
They claim to be the "Lodash replacement" but it's still a bit "bleeding edge" 😄

PR #307, opened 2024-09-01
Fix search palette tag removal and various other fixes

michaelrambeau:
## Goal

Fix various defects in the web UI:

- Number of stars by day showing weirdly, when selecting the filter "By stars added last month" and when numbers are long. E.g `+123.1 stars / day` is split after the slash character
- Removing a tag from the search palette does not work as expected (see "how to test" below)
- Upgrade [cmd-k](https://github.com/pacocoursey/cmdk) package to v1 and fix warnings in the console
- Fix missing links in the "hover card" displayed when the mouse "hovers" tags

## How to test

- Visit a "tag" page. E.g. "Component library": `/projects?tags=component`
- Open the search palette: the tag should be selected in the modal
- Click to remove the tag
- Search for another tag
- Check the page for the selected tag opens correctly

#### Search palette

image

#### Projects page / Before

image

#### Project page / After

image

#### Hover card

image

### DIFF
--- a/apps/bestofjs-nextjs/package.json
+++ b/apps/bestofjs-nextjs/package.json
@@ -48,7 +48,7 @@
     "@vercel/analytics": "^1.0.1",
     "autoprefixer": "^10.4.14",
     "clsx": "^2.0.0",
-    "cmdk": "^0.2.0",
+    "cmdk": "^1.0.0",
     "cross-env": "^7.0.3",
     "dayjs": "^1.11.8",
     "debug": "^4.3.4",
--- a/apps/bestofjs-nextjs/src/app/backend-search-requests.tsx
+++ b/apps/bestofjs-nextjs/src/app/backend-search-requests.tsx
@@ -14,7 +14,9 @@ export function getHotProjectsRequest(count = 5) {
 
 export function getLatestProjects() {
   return {
-    sort: {}, // TODO sort by `addedAt` when it's available in data instead of relying on the default sort
+    sort: {
+      added_at: -1,
+    },
     limit: 5,
   };
 }
--- a/apps/bestofjs-nextjs/src/components/project-list/sort-order-options.ts
+++ b/apps/bestofjs-nextjs/src/components/project-list/sort-order-options.ts
@@ -70,7 +70,7 @@ export const sortOrderOptions: SortOption[] = [
   {
     key: "newest",
     label: "By date of addition on Best of JS",
-    sort: {},
+    sort: { added_at: -1 },
   },
   // {
   //   key: "bookmark",
--- a/apps/bestofjs-nextjs/src/components/search-palette/search-palette.tsx
+++ b/apps/bestofjs-nextjs/src/components/search-palette/search-palette.tsx
@@ -29,11 +29,7 @@ import {
   TagSearchResult,
   ViewAllTagsCommand,
 } from "./search-palette-results";
-import {
-  SelectedItem,
-  useSearchPaletteState,
-  useSearchPaletteTags,
-} from "./search-palette-state";
+import { SelectedItem, useSearchPaletteState } from "./search-palette-state";
 import { SearchTrigger } from "./search-trigger";
 
 type CombinedSearchResult =
@@ -46,6 +42,8 @@ const TEXT_QUERY_MIN_LENGTH = 2;
 
 export function SearchPalette() {
   const {
+    currentTags,
+    currentTagCodes,
     debouncedOnChange,
     isPending,
     onOpenChange,
@@ -54,11 +52,11 @@ export function SearchPalette() {
     onSelectSearchForText,
     onViewAllTags,
     open,
+    removeTag,
     searchQuery,
     selectedItem,
     setOpen,
   } = useSearchPaletteState();
-  const { currentTags, currentTagCodes, removeTag } = useSearchPaletteTags();
 
   const canTriggerSearch = searchQuery.length >= TEXT_QUERY_MIN_LENGTH; // we need at least 2 characters to trigger a search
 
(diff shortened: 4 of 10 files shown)


apps/bestofjs-nextjs/package.json
```diff
@@ -48,7 +48,7 @@
     "@vercel/analytics": "^1.0.1",
     "autoprefixer": "^10.4.14",
     "clsx": "^2.0.0",
-    "cmdk": "^0.2.0",
+    "cmdk": "^1.0.0",
```
michaelrambeau, line 51: Changes needed:

- `aria-selected:` => `aria-[selected='true']:`
- `data-[disabled]:` => `data-[disabled='true']:`

apps/bestofjs-nextjs/src/components/ui/command.tsx
```diff
@@ -28,7 +28,11 @@ Command.displayName = CommandPrimitive.displayName;
 const CommandDialog = ({ children, ...props }: DialogProps) => {
   return (
     <Dialog {...props}>
-      <DialogContent className="overflow-hidden p-0 shadow-2xl xl:max-w-[800px]">
+      <DialogContent
+        className="overflow-hidden p-0 shadow-2xl xl:max-w-[800px]"
+        aria-describedby={undefined}
```
zawarudo, line 33: 😨 
This is scary right?

Whats going on passing `undefined` to `aria-describedby`?

apps/bestofjs-nextjs/src/app/backend-search-requests.tsx
```diff
@@ -14,7 +14,9 @@ export function getHotProjectsRequest(count = 5) {
 
 export function getLatestProjects() {
   return {
-    sort: {}, // TODO sort by `addedAt` when it's available in data instead of relying on the default sort
+    sort: {
+      added_at: -1,
```
zawarudo, line 18: I'm not understanding this line.

I'd like to spend more time to dig into this but I cannot.

So I will start the discussion around my understanding as a "new dev" here.


You encapsulate the queries in `backend-search-requests.tsx` (Not .ts?)

These are basically configuration objects that will be consumed by your ORM right?

I'm just a bit worried that the values are not self explanatory, something like `limit: 5` is easy enough to understand without having to find some extra documentation, but for now the sort configuration looks like black magic to me.

apps/bestofjs-nextjs/src/components/project-list/sort-order-options.ts
```diff
@@ -70,7 +70,7 @@ export const sortOrderOptions: SortOption[] = [
   {
     key: "newest",
     label: "By date of addition on Best of JS",
-    sort: {},
+    sort: { added_at: -1 },
```
zawarudo, line 73: Same discussion as my previous comment regarding `added_at: -1`

apps/bestofjs-nextjs/src/components/search-palette/search-palette.tsx
```diff
@@ -54,11 +52,11 @@ export function SearchPalette() {
     onSelectSearchForText,
     onViewAllTags,
     open,
+    removeTag,
     searchQuery,
     selectedItem,
     setOpen,
   } = useSearchPaletteState();
-  const { currentTags, currentTagCodes, removeTag } = useSearchPaletteTags();
```
zawarudo, line 61: So if I'm reading this correctly.

`useSearchPaletteTags` is removed and its functionality is included in `useSearchPaletteState`?

As I don't see any changes to both hooks in this PR the situation is unclear.

michaelrambeau: The problem: before was we were calling 2 different hooks `useSearchPaletteState()` and `useSearchPaletteTags()` that in turn was calling `useSearchPaletteState()`, so the result was the two calls of `useSearchPaletteState()` when rendering, which created two different states and caused the bug (because there was not context or any mechanism to share the state involved, it's a common error when composing hooks!)

This is why I chose to remove  the hook `useSearchPaletteTags` and only use what is returned by `useSearchPaletteState` instead.

apps/bestofjs-nextjs/src/app/backend-search-requests.tsx
michaelrambeau: What we call backend in the Next.js app is the big JSON file that contains aggregated data (including the trends), rebuilt every day. The Postgres DB is used only when opening the project "details" page, not to generate lists of projects.

We query this file using MongoDB-format queries (using Mingo package) when showing any list of projects on the screen.

I chose this solution because MongDB queries are just plain JS objects that are very convenient to express the criteria, the sort order, the number of projects we want.

So this file `backend-search-request` only contains two queries that I used in different places, to keep the code dry. It should be `.ts`, not `.tsx` 👍 

The following means: the MongoDB query to get the 5 first projects by `added_at` date.

```ts
export function getLatestProjects() {
  return {
    sort: {
      added_at: -1,
    },
    limit: 5,
  };
}
```

apps/bestofjs-nextjs/src/components/ui/command.tsx
michaelrambeau: Here is the exact warning message:

image

I chose `aria-describedby={undefined}` because the Command palette already has this, for accessibility:

```
<DialogTitle className="sr-only">Search</DialogTitle>
```

zawarudo: Oh, I learned something new today thanks.

apps/bestofjs-nextjs/src/components/search-palette/search-palette.tsx
zawarudo: Makes sense!

PR #313, opened 2024-09-15
Trigger E2E tests when Vercel Preview is ready

michaelrambeau:
## Goal

Make PR checks faster by moving the E2E tests to a separate workflow, triggered when deployment on Vercel is ready, following the approach described here: https://vercel.com/guides/how-can-i-run-end-to-end-tests-after-my-vercel-preview-deployment


michaelrambeau:
Another approach was set here: https://github.com/bestofjs/bestofjs/pull/315

PR #332, opened 2024-11-27
Refactor backend tasks and split schema

michaelrambeau:
## Goal

- Split the big shema file into separate files (basically one by table, excluding relation tables)
- Simplify tasks, using where clauses introducing here #331  instead of `shouldProcess` functions to filter the projects to process

Example: `{ where: notInArray(schema.projects.status, ["deprecated", "hidden"]) }`

## How to test

Ensure tasks such as `build-static-api` only process projects that are not `deprecated`.

```sh
 bun run apps/backend/src/cli.ts build-static-api 
```


apps/backend/src/iteration-helpers/repo-processor.ts
```diff
@@ -28,7 +28,8 @@ export class RepoProcessor extends ItemProcessor<Repo> {
       .from(repos)
       .orderBy(desc(repos.added_at))
       .offset(skip)
-      .leftJoin(projects, eq(projects.repoId, repos.id));
+      .leftJoin(projects, eq(projects.repoId, repos.id))
+      .groupBy(repos.id);
```
michaelrambeau, line 32: Needed because repos can be linked to multiple projects, we want each repo to be processed only once!

PR #249, opened 2023-11-14
Use Klass instead of CVA

flamrdevs:
## Goal

- Fixes #248 

Improve performance by replacing class variant utility library from cva to klass.

Other changes :
- Add custom tailwind intellisense.
- Refactor Button component with `klassed` API, this change polymorphic behavior from `asChild` prop to `as` prop.
- Refactor some components that use buttonVariants to remove `cn` util, since the buttonVariants is pre-configured to handle twMerge internally

## How to test

Ensure the class output is as expected

PR #352, opened 2025-04-26
Add trends/* pages and time range picker to the home page

michaelrambeau:
## Goal

This is the continuation of the work from @SaulSilver started here #274 . Because of the conflicts, I preferred doing another PR, sorry Hatem!

Goal: port a feature available in the previous Best of JS; the ability to check the trends for various time ranges directly from the top page.

So this PR adds a dropdown menu in the top page with 4 options:

- Today (default choice)
- Last 7 days →opens `/trends/weekly` page
- Last 30 days →opens `/trends/monthly` page
- Last 30 days →opens `/trends/yearly` page

Advantage of this approach: static generation of the new pages, that was not possible with the existing pages that rely on query string parameters.

Strategy:

The top page and the 3 new trends/* pages are grouped under the folder `(home)` to share the same `Layout.tsx` that provides all the components, except for the part related to the "hot projects".

```
(home)
├── layout.tsx
├── page.tsx
└── trends
    ├── monthly
    │   └── page.tsx
    ├── weekly
    │   └── page.tsx
    └── yearly
        └── page.tsx

```

### Other changes

Remove `@klass/*` packages introduced by #249 and use `"class-variance-authority"` because I prefer to keep components consistent with `shadcn/ui` and be able to update easily if necessary.

## How to test

Visit the 3 news pages, checking the behavior of the time range drop-down menu.

Navigation should be fast because of the static generation.

## Screenshots

image


zawarudo (approved):
Works well, n8n is killing the charts.

No issues with the code.

PR #381, opened 2025-11-02
Fix shuffle featured projects in the home page

michaelrambeau:
## Goal

Fix the behavior of the shuffled projects in the top page.
In principle, the list of projects should be shuffled and users should be able to navigate bu using the navigation buttons.
It works only when running in local but In reality, it the projects are shuffled almost everytime the user navigates.

Let's fix this by storing in the database the order of the projects to be displayed in the "Featured" panel, introducing a new table called `daily_featured_projects` that stores the list of slugs.

The daily script that builds the static API will create a new record everyday.

Also included:

- Avoid layout shifts when navigating through the featured projects panel (fixing a minimal height)
- Avoid errors if projects by slug are not found (which can happen if the `daily_featured_projects` record contains a slug that was updated

## How to test

To generate the daily data:

```sh
bun run apps/backend/src/cli.ts build-static-api                                                                                   
```

The first record created in the DB:

image

## Screenshots

image

### DIFF
--- a/apps/backend/src/tasks/build-static-api.task.ts
+++ b/apps/backend/src/tasks/build-static-api.task.ts
@@ -204,7 +204,6 @@ export async function buildFeaturedShuffledProjects() {
 function dateToDailyRecordKey(input: Date) {
   const date = normalizeDate(input);
   return [date.year, date.month, date.day]
-    .map(String)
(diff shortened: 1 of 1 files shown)


michaelrambeau:
@codex review

PR #383, opened 2025-11-07
Fix typewriter effect when leaving the page and going back

michaelrambeau:
## Goal

After the migration to Next.js 16 #380 the typewriter effect behave oddly when leaving the home page and going back!

From Cursor:

### Bug
- **What happens**: The typewriter runs an async `while` loop that sleeps via `setTimeout` and mutates non-reactive locals. It doesn’t cancel when the component is hidden/unmounted, so it continues (throttled) in the background and/or restarts, leading to overlapping loops and desynchronized state. Returning to the page makes the effect “catch up” and type frantically or stop, depending on timing.
- **Root causes**:
  - No cancellation/cleanup for the long-lived async loop.
  - Loop started via a delayed callback can run again on reactivation.
  - Mutable locals (`let` variables) reset on render, causing sequencing bugs with an in-flight loop.

### Why it appears with Next.js 16 (and not before)
- **Route caching and partial reactivation**: Next.js 16 keeps route trees “warm” and can restore them without a full unmount/remount. The home component may remain mounted offscreen or be quickly reactivated, so timers and async loops keep running or resume instead of being reset as in earlier versions where navigation typically unmounted the page.
- **Background timer throttling/resume**: While the page is hidden, timers are throttled; when re-shown, pending timeouts can resolve in quick succession, making the loop appear to “rush.”
- **Effect re-entry without safeguards**: With softer navigations and restored trees, the delayed starter can fire again unless explicitly guarded, creating overlapping loops—an issue that was previously masked by full unmounts.
 
### Fix
- Start once, with lifecycle control: Replace useTimeout with a useEffect that schedules the initial 3s start and guards with startedRef so the loop can’t start multiple times.
- Proper cancellation and restart: On cleanup, set cancelledRef to true, clear the pending timeout, and reset startedRef to false. On mount, reset cancelledRef to false so the loop can run again after navigation.
- Persist internal state: Move mutable locals to refs (currentTopicIndexRef, reverseRef) so state isn’t reset on re-render while the async loop is mid-flight.
- Safety checks inside the loop: Check cancelledRef between awaits in animateText, incrementText, and reduceText to bail out immediately when the component is hidden/unmounted.

Result: No overlapping loops, no “catch-up” bursts, and the typewriter reliably restarts when returning to the page under Next.js 16’s route caching/restoration.

## How to test

Leave the home page and go back after a few seconds, ensure the typewriter does not go crazy 😅

### DIFF
--- a/apps/web/src/components/home/typewriter.tsx
+++ b/apps/web/src/components/home/typewriter.tsx
@@ -11,15 +11,27 @@ type Props = {
 export function TypeWriter({ topics, sleepTime, loop = false }: Props) {
   const typeWriterRef = useRef<HTMLSpanElement | null>(null);
   const defaultTopic = topics[0];
-  let currentTopicIndex = 0;
-  let reverse = true;
+  const startedRef = useRef(false);
+  const cancelledRef = useRef(false);
+  const currentTopicIndexRef = useRef(0);
+  const reverseRef = useRef(true);
 
   useEffect(() => {
+    // When (re)mounted or made visible again, allow running and clear cancellation
+    cancelledRef.current = false;
+    if (startedRef.current) return;
+    startedRef.current = true;
+
     const timeout = setTimeout(() => {
-      animateText();
+      if (!cancelledRef.current) {
+        animateText();
+      }
     }, 3000);
 
     return () => {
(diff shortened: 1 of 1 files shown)


michaelrambeau:
@codex review

Issue #337, opened 2025-01-14
Project tracking proposal: Base UI


oliviertassinari:
I would propose bestofjs tracks Base UI

- Stars history: https://star-history.com/#mui/base-ui&Date
- First alpha introduction: https://x.com/colmtuite/status/1869053012712550819

I'm not sure if this is done automatically base on projects trending on GitHub or if it's a manual process, so in the doubt

michaelrambeau:
Hello Olivier @oliviertassinari , it's nice to see you again here 😄 !

The initial tracking is manual, as all projects in Best of JS were added and tagged manually by myself.

`Base UI` will be available on _Best of JS_ very soon.

michaelrambeau:
Hello @oliviertassinari `Base UI` was added to _Best of JS_, please check:

https://bestofjs.org/projects/base-ui

oliviertassinari:
@michaelrambeau Sweet, thanks.

For the logo @colmtuite, I'm not sure exactly what to provide from https://www.notion.so/mui-org/Branding-book-42011f0b6ad44e7ba4e507c00b2da13a?pvs=4#171cbfe7b660801d8664e451c1879094. But I guess this logo is a safe starting point https://x.com/base_ui (X is round, Best of JS is a square, so centering is likely different)

michaelrambeau:
For the logo, I only need the square SVG file (or 2 of them to handle the dark mode) but I can grab them from the documentation page, no problem. I will do that soon, I will confirm when it's online.

FYI I think the Notion page you shared is private.

michaelrambeau:
Hello @oliviertassinari please check again, I have added the logo for light and dark modes. https://github.com/bestofjs/bestofjs/commit/5d3793229123be7c129b2995048947f42c081788

https://bestofjs.org/projects/base-ui

![Image](https://github.com/user-attachments/assets/16632a13-4890-419d-aacc-8971e2585630)

colmtuite:
Thanks @michaelrambeau 🙏

michaelrambeau:
@oliviertassinari Hello again Oliver forget about my previous question, I could reproduce the issue.

Yesterday I switched to light mode and there was no issue.
Today, the home page was re-built and I checked the site (like I do every morning 😅 ) and I noticed the issue.
So the issue happens when you are in light mode, before you use the color mode switcher, I will investigate.

PR #394, opened 2025-12-16
fix image mode loading

oliviertassinari:
Solve https://github.com/bestofjs/bestofjs/issues/337#issuecomment-3657860293, the problem can be reproduce having the light mode by default and browsing for the first time, you will see this:

https://github.com/user-attachments/assets/3bf9eb11-f7d7-4a96-8a82-18169742b9db

Why it's happening is explained in https://github.com/pacocoursey/next-themes?tab=readme-ov-file#avoid-hydration-mismatch. The simple version is that on the server side, when the page is SSR, the server doesn't know what the mode client side.

There are two different ways to solve this, documented on that same next-themes page:

1. By waiting for the page de be hydrated:

```diff
diff --git a/apps/web/src/components/core/project-logo.tsx b/apps/web/src/components/core/project-logo.tsx
index 6758760a..1404778e 100644
--- a/apps/web/src/components/core/project-logo.tsx
+++ b/apps/web/src/components/core/project-logo.tsx
@@ -1,5 +1,6 @@
 "use client";

+import { useState, useEffect } from 'react';
 import Image from "next/image";
 import { useTheme } from "next-themes";

@@ -67,8 +68,20 @@ export function ProjectCustomLogo({
 }: {
   filename: string;
 } & BaseProps) {
-  const colorMode = useColorMode();
-  const imageURL = getProjectCustomLogoURL(filename, colorMode);
+  const { resolvedTheme: colorMode } = useTheme();
+  const [mounted, setMounted] = useState(false);
+  let imageURL;
+
+  useEffect(() => {
+    setMounted(true);
+  }, []);
+
+  if (mounted) {
+    imageURL = getProjectCustomLogoURL(filename, colorMode!);
+  } else {
+    imageURL = 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7';
+  }
+
   return (
     <Image
       src={imageURL}
@@ -99,8 +112,3 @@ export function GitHubAvatar({
     />
   );
 }
-
-function useColorMode() {
-  const { resolvedTheme } = useTheme();
-  return (resolvedTheme || "dark") as "dark" | "light";
-}
```

2. Or by having two images, toggling them with CSS. It's the strategy use on https://nextjs.org/. It's better because it allows to load the image asset a lot sooner.

### DIFF
--- a/apps/admin/src/components/project-logo.tsx
+++ b/apps/admin/src/components/project-logo.tsx
@@ -74,7 +74,7 @@ export function getProjectLogoUrl(
   return url;
 }
 
-function getProjectLogoURL(input: string, colorMode: string) {
+function getProjectLogoURL(input: string, colorMode: "dark" | "light") {
   const [main, extension] = input.split(".");
   const filename = colorMode === "dark" ? `${main}.dark.${extension}` : input;
   return `https://bestofjs.org/logos/${filename}`;
--- a/apps/web/src/app/globals.css
+++ b/apps/web/src/app/globals.css
@@ -163,3 +163,8 @@ footer::after {
     #9c0042 80%
   );
 }
+
+.dark .light-mode-only,
+.light .dark-mode-only {
+  display: none;
+}
--- a/apps/web/src/components/core/project-logo.tsx
+++ b/apps/web/src/components/core/project-logo.tsx
@@ -1,7 +1,7 @@
 "use client";
 
+import { Fragment } from "react";
 import Image from "next/image";
-import { useTheme } from "next-themes";
 
 import { cn } from "@/lib/utils";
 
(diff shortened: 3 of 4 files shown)


michaelrambeau:
Thank you very much Olivier, I have seen you are working on that right now, I was checking the issue too.

The agent was suggesting another approach:

> You need to prevent rendering the theme-dependent logo until the theme is actually loaded. The next-themes library provides a mounted state for this

```tsx
function useColorMode() {
  const { resolvedTheme } = useTheme();
  const [mounted, setMounted] = useState(false);

  useEffect(() => {
    setMounted(true);
  }, []);

  // Prevent hydration mismatch by returning undefined until mounted
  if (!mounted) {
    return undefined;
  }

  return (resolvedTheme || "dark") as "dark" | "light";
}
```

That would require a combination of `useState` and `useEffect`, so I think your solution is better, the only downside being that all custom logos would be downloaded twice by the browser but they are usually quite small (around 5 kB).

Preview URL is available here: https://bestofjs-git-fork-oliviertassinari-fix-light-mo-66d40e-bestofjs.vercel.app/

oliviertassinari:
> The agent was suggesting another approach:

Right, it's option 1. from above.

michaelrambeau (approved):
Thank you so much for the fix and the explanations Oliver 🎉 , let's merge and I will deploy to production.

Issue #459, opened 2026-07-25
Add options to remove AI related projects from listings while viewing


L1lith:
It would be cool to have a simple toggle to permanently remove all AI related projects from the site while you are browsing. This is important because some people vehemently do not want to see AI related things under any circumstances. I think it would be good to allow them to have a better experience while still allowing people who want to see AI related projects to see them. It would also be good to allow people to fine tune specifically what kind of AI related projects they want to filter out, for example is a project coded with AI allowed even if it's purpose isn't related to AI? I also think it would be cool because it would require projects to disclose their usage of AI, perhaps through a statement and a few checkboxes. Again I am not advocating against AI here I am just advocating for transparency and putting the power in the hands of the people using BestOfJS. I think this has the potential to create good will among people who don't like AI and improve their experience dramatically.

Also it would be cool to add a "noai.bestofjs.org" subdomain which would have strict noai filtering which would allow people to find the no-ai version of the site directly from google. Duckduckgo already does this and has received a lot of praise for it.

michaelrambeau:
Hello @L1lith long time no see 😄 

Great idea, my main concern about AI projects is that AI related projects are an order-of-magnitude more popular than other projects so they tend to make Best of JS rankings very boring, the same projects show up almost every day:

- last year n8n was the most popular project almost every single day
- nowadays it's about project like Matt's skills, OpenCode or Pi...

I use DuckDuckGo but I was not aware of that setting actually (screenshot below)

Is it what you mean, when you said "no AI" version?

<img width="797" height="851" alt="Image" src="https://github.com/user-attachments/assets/0edcec46-7030-4498-b3a7-a5c17e5ee8e8" />

Doing a lot of refactor related to Best of JS architecture currently (see #371 ) I will try to think about that after i finish the migration.

L1lith:
Good to talk to you again as well @michaelrambeau 😄 What you're saying is mostly correct. I think there will be significantly less projects for sure but I still think it's a great thing to offer. What you're suggesting is correct for the main domain, but for the subdomain it would correlate more to this site: [https://noai.duckduckgo.com](https://noai.duckduckgo.com), which is a mirror of the main site except AI is always removed to the maximum degree even without needing to configure anything.

[Personal Note]
[...]

PR #505, opened 2026-09-14
Update APP_DISPLAY_NAME for no ai version

BBlackwo:
## Goal

Update title for No AI version to `Best of JS (No AI)`. Follows on from https://github.com/bestofjs/bestofjs/issues/459.

### NO AI

Screenshot 2026-09-14 at 10 11 37 pm

### Normal

Screenshot 2026-09-14 at 10 12 13 pm


michaelrambeau:
Thanks you Benjamin!

Sorry for the delay, I suggest the following approach, if it makes sense to you too?

---

Right change, but I'd like the app name to live in the registry rather than a second `process.env` read.

We already have `apps/web/src/config/apps.ts`, which maps `BESTOFJS_APP` to everything that differs between the two deployments (excluded tags, header badge, host). The `"noai"` literal lives there and in `env.mjs`; this adds a third place that knows about it.

There's also a subtler issue: `config/site.ts` is imported by client components (`mobile-nav.tsx`, `desktop-nav.tsx`, `app/projects/error.tsx`). `BESTOFJS_APP` is a server-only variable, so in the client bundle `process.env.BESTOFJS_APP` is `undefined` and the ternary quietly resolves to `"Best of JS"`. Nothing renders it client-side today, so it's not a visible bug — but it's a constant that's wrong in half the bundle, waiting for someone to use it in a client component.

Note that we can't fix that by having `site.ts` import `@/config/apps`: `apps.ts` reads `env.BESTOFJS_APP` at module scope, so pulling it into the client bundle throws t3-env's "attempted to access a server-side environment variable on the client". That's the same constraint behind the badge being passed down as a slot in `site-header.tsx`.

So the shape I'd suggest — add to `config/apps.ts`, alongside the other per-app tables:

```ts
export const displayNameByApp: Record<WebApp, string> = {
  [webApps.MAIN]: "Best of JS",
  [webApps.NOAI]: "Best of JS (No AI)",
};

export const appDisplayName = displayNameByApp[currentApp];
```

Then drop `APP_DISPLAY_NAME` from `config/site.ts` and point its consumers at `appDisplayName`. All eight are server components, so the import is safe:

- `app/layout.tsx`
- `app/about/page.tsx`
- `app/projects/page.tsx`
- `app/projects/[slug]/page.tsx`
- `app/rankings/monthly/[year]/[month]/page.tsx`
- `app/hall-of-fame/layout.tsx`
- `components/footer/footer.tsx`
- `components/home/home-sections.tsx`

Please remove it rather than re-exporting from `site.ts` — a re-export would drag `env.mjs` into the client bundle and break the nav.

Last thing: `docs/features/noai-app.md` currently lists the public name as `Best of JS — No AI`. I prefer your parenthesised version (reads better in the `${APP_DISPLAY_NAME} • %s` title template), so let's keep it and update that line in the doc as part of this PR.

BBlackwo:
Hi Michael @michaelrambeau,

Makes sense 👍 Seems like a more robust solution.

